PR #78, opened 2026-08-08
RC aliasing Stage 1: mutation summaries reclaim the +23%, plus two profiled perf levers

shd101wyy:
## Summary

Completes the compile-time RC arc opened by #77: **aliasing Stage 1** (per-callee mutation summaries) lands in both compilers and **returns the entire +23% self-compile cost Stage 0 charged** — the borrowed-argument aliasing hole is now closed at no measured runtime cost. Two further perf levers found by re-profiling take the self-compile ~21% below where it started.

`check ./std` under the self-hosted binary, min user of 3:

| stage | min user |
| --- | --- |
| pre-Stage-0 | 45.3 s |
| Stage 0 (#77) | 55.2 s (+23%) |
| + Stage 1 | 45.1 s |
| + intern HashSet | 37.1 s |
| + bulk `write_string` | 35.7 s |
| **+ soundness/hardening fixes (final)** | **38.5 s** |

The last row is the honest cost of correctness — the fixed walk visits statements it had been wrongly skipping, so more callees are correctly classified as may-mutate and keep their dup. **Net: 55.2 s → 38.5 s (−30%) against the Stage-0 baseline this builds on, and 45.3 s → 38.5 s (−15%) against the pre-aliasing baseline — with the hole closed.**

## Stage 1 — mutation summaries

Stage 0 gives every RC-typed field **projection** passed to a **borrowing** parameter a caller-owned `+1`, because the callee might reassign that storage through an aliased handle mid-call. Most callees never do. Stage 1 asks per callee — *"may this call transitively mutate RC container storage?"* — and gives the read-only majority their `+0` borrow back.

The walk (`src/evaluator/effects/mutation-summary.ts`, ported to `yo-self/evaluator/effects/mutation_summary.yo`) runs over **evaluated** bodies with callees resolved through the call node, memoized by funcId with least-fixpoint cycle handling: a verdict that depended on an in-progress walk is not memoized unless it found a real mutation.

**May-mutate sources:** RC-typed place assignments; externs that are runtime/allocator family (`__yo_`-prefixed), libc `free`/`realloc`, or take callbacks; io builtins (they yield to the event loop); ctl functions; any unresolved callee. Raw C byte operations cannot decrement a reference count, so `memcmp`/`printf`/libc math stay read-only. Match-arm *patterns* are skipped (Yo has no guards, and variant constructors have no resolvable callee); primitive builtin heads (`__yo_op_*`, pointer math, `rc`, `__yo_panic`) are whitelisted.

**Gating:** non-generic callees skip the Stage-0 mark outright; generic callees are marked and unmarked after specialization (`removeBorrowedProjectionDupMark` reverses the marker exactly — dup removed, temp flipped non-owning). Recursive short-circuit calls and checking-phase trials keep their dups.

### yo-self has THREE arg-marking call paths, not two

Instrumentation turned up a path neither Stage 0 nor I knew about: **`_evaluate_funcval_runtime_call`** — where most ordinary calls actually land. It early-returns *before* the inline path's drop wiring, so Stage-0 dups there had been orphaning their balancing drop since #77 (invisible to rc-balance asserts, since the callee's own reassignment consumes a ref). Stage 1's decision now runs there too, after the Gap-6 specialization (a generic callee's raw body is unevaluated — only the minted spec's body is walkable), eliding the dup entirely for read-only callees.

That path also never wired the balancing drop for may-mutate callees — a leak dating to #77 — which is closed here too. See **Hardening** below.

## Perf levers (plans/PERF_BORROW_ELISION.md)

The plan's 2026-07 attribution table was **stale** — its lever 1 (`== String.from(x)`, "177 sites") had already landed, leaving 29 cold `asm.yo` sites. A live `sample` profile of the running compiler showed the hotspot had moved entirely, and found two wins the table had no entry for:

1. **Type-interning cycle guard, linear scan → hash set** (`yo-self/types/intern.yo`) — `_intern_key_into`'s `visited` set was an `ArrayList(String)` probed by linear scan: O(visited) full string comparisons per named type on every key render, and `substitute` re-keys the whole subtree at each recursion level. It was the **top of the profile** (`_contains` 3731 + `memcmp` 1562, ahead of `__yo_decr_rc` at 3099). Swapped for `HashSet(String)`; **-17.8%**, and `__yo_decr_rc` itself fell 3099 → 1373 samples because the probes *were* the RC traffic.

2. **`StringBuilder.write_string` per-byte push → bulk copy** (`std/string/string_builder.yo`) — it appended byte-at-a-time through `ArrayList(u8).push` while its sibling `write_str` already bulk-copied. **-4.2%** (interleaved A/B, 5 reps; 4 of 5 runs beat every baseline run).

The plan is updated with a staleness banner, the `sample` method that re-derives attribution, both landed levers, and a re-scoped lever 2 (`_attach_early_return_only_drop_to_returns` is now 1.3%, not 13.9%, but its O(vars × tree) walk and its generate-before-checking still deserve the algorithmic fix).

## Hardening: four use-after-frees found and fixed before merge

A review pass plus a systematic probe of **every** "this callee is read-only" rule found four real UAFs that the entire gate battery (fast suite, 155-program corpus, `check ./std`, fixpoint, Guard Malloc) reported green over. That is structural, not luck: those gates verify the behaviour of emitted programs, so a mis-classification is invisible unless the corpus happens to contain the triggering shape.

1. **Compile-time-value early-exit outranked the structural forms.** Every statement carries a compile-time value (an assignment is unit-typed, a declaration has its own), so the walk skipped whole statements and never saw a mutating call on the right-hand side. `t := poke(w, seed)` was reported read-only; the borrow read recycled memory, **returning 2 instead of 42**.
2. **Every RC-DECREMENT primitive was whitelisted as safe** — `___drop`/`___dispose` in the safe-recurse list, `__yo_decr_rc`/`__yo_dyn_drop`/`__yo_sometype_drop` in the pure list, while the array/tuple element drops were correctly excluded. A callee doing `___drop(w.b)` releases the container's reference with no assignment to see. Decrements now have their own set, tested *before* the whitelists.
3. **The memo cache was keyed by a reusable id.** `funcId`s are per-module counters that the test/build runners reset between compilations, so a stale "read-only" verdict could be inherited by an unrelated mutating function.
4. **The unmark was fail-open** — it could remove a dup while leaving its temp owning, handing the RAII pass a drop with no matching dup.

**Rules probed and found sound:** cycle-guard optimism on a mutual-recursion back edge, invoked closures, `while` bodies, live `cond`/`match` arms, three-deep call transitivity, and the post-fix compile-time-value exit. Compiler-*generated* drops are provably not a hazard: they never appear as AST children, and `getVariablesNeedingDrop` only selects owning variables — a borrow is non-owning by definition.

**Known leak closed.** `_evaluate_funcval_runtime_call` — where most ordinary calls land — never wired its balancing drop, so every may-mutate Stage-0 dup on that path leaked its `+1` from the day Stage 0 shipped.

**Release points reconciled.** Closing that leak made a pre-existing design difference observable: yo-self releases the `+1` post-call, TS at scope end. Both sound, observable only through `rc`; the leaking path had been matching TS by *never releasing at all*. The regression test now asserts after the enclosing scope, where both agree (`rc == 1`) — which catches a leak and a double-release equally. Notably the previous assertion (`rc == 2` right after the call) had encoded the leak as expected behaviour. Unifying the two release points is left as its own change: TS emits call-node drops *before* the call in some emitters, so moving its release risks the UAF already fixed in yo-self's `recur.yo`.

## Validation

| Gate | Result |
| --- | --- |
| `tests/rc.test.yo` (11 new tests: elision on all three call paths + one adversarial probe per decision rule + an over-conservatism guard) | **32/32 under BOTH compilers**; each soundness probe verified RED before its fix |
| TS fast suite (`./tests` minus internal) | **2673 passed / 0 failed** |
| Self-hosted battery (23 files, hollow detection) | 0 failures, hollow=0 |
| Differential corpus | **155 PASS / 0 DIFF** |
| `yo-self-bin check ./std` | 153/153 |
| Stage-2 → clang → stage-3 | **FIXPOINT HOLDS**, hollow=0 |
| Emission-neutrality of the intern change | **25 corpus programs emit BYTE-IDENTICAL C** pre/post |
| Perf guard | elision rate held after the broader decrement rule (38.5 s, not a regression toward 55 s) |
| Guard Malloc (repro) | clean |

Each of the three landed changes was gated independently (battery + corpus + `check ./std` + fixpoint per change).

## Docs

`docs/en-US/COMPILE_TIME_RC_WITH_OWNERSHIP_ANALYSIS.md` and its zh-CN twin: the **"Known aliasing gap"** warning becomes a closed item describing both stages and the measured cost.


shd101wyy:
## Review pass: two soundness holes found and fixed (plus two hardening fixes)

A code review of this branch turned up a **real use-after-free** in the Stage-1 walk that every gate I had run was blind to. Both compilers were affected. Fixed in `de5cd8f9` and `83393a0f`.

### 1. Compile-time-value early-exit ran above the structural forms (HIGH — UAF)

The walk short-circuited on `info.value && !isUnknownValue(...)`, reasoning "this was folded away, so nothing inside runs". But **every statement-shaped expression carries a compile-time value** — an assignment stamps `VUnit`, a declaration carries its own — so the walk skipped whole statements and never visited a mutating call on the right-hand side:

```rust
poke   :: (fn(w : AliasWrap, seed : i32) -> i32)({ w.b = box(i32(100) + seed); w.b.* });
honest :: (fn(w : AliasWrap, borrowed : Box(i32), seed : i32) -> i32)({
  t := poke(w, seed);          // ← skipped whole; the mutation is never seen
  borrowed.* + (t - t)
});
```

`honest` was reported **read-only**, Stage 1 elided the Stage-0 dup, and the borrowed projection read recycled memory: **returned 2 instead of 42**. Moved the check below the structural dispatch, where it only applies to genuinely CTFE'd calls. Ported to yo-self in lockstep.

### 2. Memo cache keyed by a reusable id (HIGH — latent UAF)

`summaryByFuncId` is process-global, but `funcId`s come from `randomId()` — a **per-module counter** that `clearAllModuleCounters()` resets per test file and `resetModuleIdCounter()` resets on module reload. A stale "read-only" verdict could therefore be inherited by an unrelated mutating function. `utils.ts` now exposes a module-id generation (a plain number — no import cycle into that leaf); the cache drops its memo when it moves.

yo-self does **not** share this hazard: its `random_id` is a single monotonic process-global with no reset. That invariant is now documented at the cache, with a note to add the guard if the generator ever gains a reset.

### 3. Unmark was fail-open (MEDIUM)

`removeBorrowedProjectionDupMark` cleared the deferred dup *before* confirming it could flip the dup temp to non-owning. On either early-return path the dup was gone but the temp stayed owning — and the caller's RAII collection runs immediately after, so it would emit a scope-end `___drop` with no matching `___dup`. Now all-or-nothing.

### Regression test

`tests/rc.test.yo` — *"statement-hidden mutation keeps the borrowed-projection dup"* — covers hole 1 in both the reassignment and declaration forms, and was **verified red before the fix** (SIGABRT). rc.test.yo is 24/24.

Re-running the full gate chain (battery, corpus, `check ./std`, fixpoint, fast suite) on the corrected compiler; CI is re-running on this head.

shd101wyy:
## Corrected performance numbers (post-soundness-fix)

The figures in the PR body were measured **before** the soundness fixes, so they are optimistic — part of that speed came from the walk wrongly skipping statements. Re-measured on the corrected compiler (`check ./std`, self-hosted, min user of 3):

| stage | min user | vs prior |
| --- | --- | --- |
| pre-Stage-0 | 45.3 s | — |
| Stage 0 (#77) | 55.2 s | +23% (the cost this PR reclaims) |
| **+ Stage 1** | 45.1 s | −18% |
| + intern HashSet | 37.1 s | −17.8% |
| + bulk `write_string` | 35.7 s | −4.2% |
| **+ soundness fixes** | **39.0 s** | **+9.3%** |

That last row is the honest cost of correctness: the walk now visits statement bodies it had been skipping, so more callees are correctly classified as *may-mutate* and keep their Stage-0 dup. The prior 35.7 s was partly the bug.

**Net effect of this PR: 55.2 s → 39.0 s (−29%) against the Stage-0 baseline it builds on, and 45.3 s → 39.0 s (−14%) against the pre-aliasing-fix baseline — with the aliasing hole closed.** So the +23% Stage 0 charged is not merely reclaimed, the self-compile ends up meaningfully faster than before the aliasing work started.

Re-validated end-to-end on the corrected compiler: battery 0 failures / hollow=0, corpus **155 PASS / 0 DIFF**, `check ./std` 153/153, **FIXPOINT HOLDS**, `tests/rc.test.yo` 24/24 under **both** compilers.

PR #97, opened 2026-08-11
Fix the native-Windows self-hosted compiler build (v0.2.0 seed-leg porting tail)

shd101wyy:
Fixes `issues/fixed/windows-native-selfhosted-build-fails.md` — the three error classes from the v0.2.0 windows-x64 seed leg, each at its source:

- **`F_OK`**: `_file_exists_sync` (import resolver) now uses portable libc `stat` (MSVC-clean, mirrors TS `existsSync`) instead of POSIX `access`.
- **`setenv`**: the self-hosted test runner sets `YO_TEST_INDEX` via `std/env`'s portable `env.set` (Windows `_putenv_s` branch).
- **main-wrapper returns**: the TS Windows/WASM wrapper now emits module-level init into the dedicated `void __yo_main_module_init` helper like POSIX (effectful initializers' bare `return;` escape checks were landing inside `int main`). yo-self already had the helper but lacked the post-init `__yo_effect_escaped` checks on both paths — mirrored.

Verified: a full Windows cross-emit of `yo-self/main.yo` greps **0** for all three classes (3 remaining `Unknown type:` markers are the benign comptime-only comment class present in working binaries); macOS suite 162/162, cli-diff corpus 27/27, FIXPOINT_HOLDS. The definitive proof is the windows CI native-build step (arrives via PR #95) going green — then it flips gating and the release windows-x64 leg gets promoted.

### DIFF
--- a/issues/fixed/windows-native-selfhosted-build-fails.md
+++ b/issues/fixed/windows-native-selfhosted-build-fails.md
@@ -1,0 +1,24 @@
+## FIXED (2026-08-11, branch fix/windows-native-selfhosted)
+
+All three classes, at their sources:
+
+1. `F_OK` — `yo-self/evaluator/exprs/import.yo` `_file_exists_sync` now uses
+   portable libc `stat` (MSVC-clean; mirrors TS `existsSync` and
+   `module_manager.yo`'s `_path_exists_sync`) instead of POSIX `access`.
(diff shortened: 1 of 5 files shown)


shd101wyy:
Merged into PR #95 (user request) so the windows native-build CI step can gate with the fix in the same PR — see the step's smoke-run there. Branch kept for history.

PR #98, opened 2026-08-11
P2.3: the seed (previous release) builds stage-1 in every self-hosted CI job

shd101wyy:
First slice of the 2.3 CI migration (design in `plans/P2_RETIRE_SRC.md` §2.3): a new `install-seed` composite action downloads the pinned release bundle (self-locating, no YO_STD) and the four self-hosted jobs switch their stage-1 builder from the TS compiler to the seed — bootstrap fixpoint and hollow sweep + internal differential drop bun/node entirely; tier-1 gates keep bun only for the TS corpus ground-truth arm. The trust chain becomes **previous release builds current compiler**.

Kept for later: TS shards/unit tests/wasm/TSan keep bun until 2.5; the per-leg native probe keeps the TS builder (a seed self-build needs ~9-11.5 GB — arm64 macOS runners have 7 GB). `SEED_VERSION` pins v0.2.0 and bumps each release (v0.2.1 will add windows-x64/macos-x64 bundles).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #112, opened 2026-08-12
Release mechanics: bump ahead of the release — no PR, no CI burned

shd101wyy:
Cutting a release becomes two dispatches: **Prepare release** (bump) then **Release** (tag + ship). The bump is pushed straight to `develop` and costs no CI.

## What actually protects `develop`

The comment in the old workflow guessed wrong. There is **no classic branch protection** (`/branches/develop/protection` → 404). It is a repository **ruleset** (`branch_protection`, id 13548862) on `~DEFAULT_BRANCH` with three rules: block deletion, block non-fast-forward, and **require 15 status checks (strict)**. There is **no "require a pull request" rule** — so what rejected the bot's push is the status-check rule (a brand-new commit has none of the 15), and `bypass_actors` is exactly `RepositoryRole 5` (admin, mode `always`).

- Scoping a ruleset to "except this workflow" is **not possible** — conditions are ref-name patterns only. The only exemption axis is *who pushes*.
- So a direct push needs an **admin identity, not a ruleset change**. `[skip ci]` does not help a bot push either: the rejection is for having no *passing* checks, so the admin bypass is the entire reason a PAT is needed.
- **The ruleset is deliberately left alone.** Dropping `required_status_checks` would let `GITHUB_TOKEN` push and need no secret, but those 15 contexts are what keep develop green, what make this workflow's gate mean anything, and what currently hold #98's red tier-1 job back from merging.

## Why the bump had to move ahead of the release

Bumping inside the release job produced two defects a direct push would not fix:

1. **The tag named the previous version.** `action-gh-release` defaults `target_commitish` to `github.sha` — the pre-bump commit — so the `v0.2.2` tag points at a tree saying `0.2.1`, and `v0.2.4`'s at `0.2.3`.
2. **The shipped bundles came from a commit no CI run had covered.** The green-gate checked SHA_A; the bump created SHA_B in-job; `seed-bundles` checked out SHA_B. Not inert — `yo-self/version.yo` is *compiled into* every bundled binary.

Neither is fixable in place: nothing can have tested a commit the release itself creates.

## The shape

- **`prepare-release.yml`** (new, 1 node-only job): bumps all three version sources, verifies they agree and the tag is free, and pushes **directly to develop** with `[skip ci]`.
- **`release.yml`**: no bumping. Reads the version, asserts all three sources agree, refuses if the tag exists, pins `target_commitish` to the released SHA, and gates on the newest ancestor that is not version-only.

Tag == seed bundles == `CURRENT_YO_VERSION` == the vetted tree, one SHA.

## No CI on a bump — and the skip is proven, not asserted

A bump's entire delta is version strings: **3 files, 6 lines**, measured on the real bumps, and **no test in the repo asserts a version** (the only `0.2.x` occurrences are help-text examples in `scripts/install.*`). The 17-job suite would verify nothing.

The earlier attempt at this on `p2/ci-migration` deadlocked, and the reason is the lesson: it **trusted the commit subject** and paired it with a file list omitting `yo-self/version.yo`, so `NON_PKG` was never 0, the parent fallback never fired, and the gate landed on a `[skip ci]` commit with `CONCLUSION=none`. This one ignores the subject and proves the claim structurally — **every changed path must be a version file AND every changed line a version field**. The path test is a *subset*, so it survives `package.json`'s deletion in P2.5.

Second net: `seed-bundles` compiles `yo-self/main.yo` from the released SHA on five platforms and smoke-tests each binary, so a mangled `version.yo` cannot ship quietly.

One asymmetry: the PR fallback **amends `[skip ci]` out**, because a PR head that skips CI never reports the 15 required checks and the ruleset would hold it unmergeable forever.

## Verification

Workflow-only, so no CI arm exercises it. Both files parse. The gate script was **extracted from the workflow and executed**:

| case | result |
|---|---|
| two stacked version-only bumps | walks back twice, gates on the real commit, rc=0 |
| ancestor's run red | rc=1, refuses |
| ancestor has no run (`none`) | rc=1, refuses |
| HEAD is a real commit | no walk-back, gates on HEAD |
| `sha` output | HEAD (what gets tagged), not the gate SHA |

The version-only predicate, against real and synthetic commits:

| commit | verdict |
|---|---|
| the 3 post-0.2.3 bumps | version-only ✓ |
| pre-0.2.3 bumps (2 files, predate the `version.yo` bump) | fails closed ✓ |
| 5 real change commits | real commit ✓ |
| bump + smuggled `README.md` edit | fails closed ✓ |
| bump + code appended to `version.yo` | fails closed ✓ |
| dependency edit under a lying `chore: bump version` subject | fails closed ✓ |

Also dry-run against the real tree: the version read gives `0.2.4` from all three sources and **correctly refuses** to re-release (tag exists); release notes skip the bump commit and quote the last real change (#107); the bump sequence in a scratch copy lands all three at `0.2.5` with `version.yo` well-formed.

## Secret

`RELEASE_PAT` **is set** (fine-grained, this repo, Contents: read/write) — so the direct-push path goes live as soon as this merges. Note `workflow_dispatch` workflows only appear in the Actions UI once they are on the default branch, so **Prepare release** does not exist as a button until then. If the token is ever unset or expires, the workflow degrades to the one-click bump PR and says so in its log.

## Note for #98

#98 carries the `[skip ci]` + gate-on-parent hunks described above. This PR replaces them, so land #98 first and resolve the overlap in favour of this branch.


shd101wyy:
Superseded by #114 per maintainer direction: no separate prepare-release.yml — release.yml itself bumps with RELEASE_PAT, pushes straight to develop, tags the bump commit, then publishes. #114 keeps this PR's two key findings (tag target_commitish, gate-on-parent for [skip ci] bumps).

PR #115, opened 2026-08-13
Def-eval swallow: nine roots fixed — sibling/forward/cross-impl dispatch and value binders now resolve at definition time

shd101wyy:
Continues the campaign PR #110 landed (`issues/def-eval-swallow-remaining-roots.md` is the full measurement record). Std-baseline swallows go **19 → 10**; every fix is a faithful port of a TS behavior yo-self was missing.

## The six commits

1. **Case-2 (generic impl) per-field provisional registration** — a later sibling's def-time trial can now resolve an EARLIER sibling (`Self.new()` before `slice_copy`): each direct colon-pair field registers its real forall-stamped FuncVal into the provisional registry as it completes, superseded in place, cleared at loop end. The static lookup consults provisional LAST (below the generic-impl fallback — outranking it is the measured imm_map regression of the reverted 2026-08-12 attempt).
2. **Trait-constructor coverage + family B** — trait-ctor inner fields get the same treatment; impl-level VALUE binders (`impl(generic(T : Type, N : usize), …)`) are re-kinded to `UnknownVal(usize)` in def-time body envs via a NAME-only side table resolved purely (every env-involved resolution measurably broke the circular-import order: 87–160/247 files).
3. **Lazy on-demand forward shells** — forward references (`slice_copy → get`) resolve through shells materialized ONLY when a trial asks for that sibling. The eager TS-shaped pre-pass measurably diverged `tests/imm_map.test.yo` even with registration disabled (map.yo's early signature instantiations poison a CTFE/instantiation order the batch main depends on).
4. **Abstract generic bindings (TS parity) + comptime params bound `UnknownVal`** — cross-impl dispatch at def time (`Clone` impl calling the inherent `len`) binds `T := T'`; comptime VALUE params carry `createUnknownValue(declaredType)` like TS (runtime params stay valueless — the dup-accounting fix is preserved).
5. **Abstract bindings scoped STRICTLY to def-time trials** — outside trials (derive validation, spec body evals) stamps feed codegen directly, and an abstract match there minted an abstract-keyed specialization the emitter skips while the call site still emits (`derive_clone_complex` 'call to undeclared function'). Bisected against per-increment binaries.
6. **Debug hooks** (`YO_DEBUG_DISPATCH`, `[abstract-spec]`) + root-#3's diagnosis: the remaining trait-branch failure is a `synthesize_types` SomeT-vs-SomeT gap, pinpointed and recorded.

## Battery (all on this exact tree)

| gate | result |
|---|---|
| hollow sweep (188 language files, self-hosted) | **188/188 GREEN** |
| `check ./std` | 154/154 |
| `check ./yo-self` | 247/247 |
| canaries (`imm_map`, `derive_clone_complex`) | 21/21, 15/15 |
| `tests/internal` (self-hosted) | running locally; CI's differential covers it |

Remaining roots (~10) and the endgame (fatal trial handler) continue in the issue doc — root #3's next move is the synthesizer gap.

### DIFF
--- a/issues/def-eval-swallow-remaining-roots.md
+++ b/issues/def-eval-swallow-remaining-roots.md
@@ -530,6 +530,41 @@ product is kept out of codegen's channels (fid recording, spec minting).
 Any further degrade must answer "which codegen channel can this abstract
 value reach?" before landing.
 
+### Root #3 (797, Clone impl) — discriminator narrowed to the TRAIT-CONSTRUCTOR field path (2026-08-13, probes on the trial-scoped build)
+
+Probe series (`scratchpad` cc\_\* files, two impls on `G(T)` where the second
+calls the first's `len`; the CALLER must return a SomeT-containing type or no
+dg-trial runs and the pass is VACUOUS — `via2/via3` returning `usize` "passed"
+that way):
+
+| second impl's shape                                     | def-time trial |
+| ------------------------------------------------------- | -------------- |
+| direct field, `self : Self`, returns `Option(usize)`    | resolves       |
+| direct field, `inout(self) : Self`, returns `Option(…)` | resolves       |
+| direct field + `where(T <: Clone)`                      | (vacuous)      |
+| `Clone(clone : …)` trait constructor                    | **swallows**   |
+| CUSTOM trait constructor (`MyT(via5 : …)`)              | **swallows**   |
+
+So cross-impl dispatch inside a TRAIT-CONSTRUCTOR field's trial still fails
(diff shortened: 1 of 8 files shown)


shd101wyy:
Local battery is now complete on this exact tree (tree hash `5202ba37`): hollow sweep **188/188 GREEN**, bootstrap **FIXPOINT_HOLDS**, `tests/internal` **868 passed / 0 transpile markers** (self-hosted, `--parallel 1`), `check ./std` 154/154, `check ./yo-self` 247/247, canaries `imm_map` 21/21 and `derive_clone_complex` 15/15.

PR #134, opened 2026-08-17
P2.5 step 24: seed-build Linux legs + cross-emit macOS legs (option A)

shd101wyy:
## What

The `seed-bundles` matrix gains a `seed_built` flag. The two Linux legs (linux-x64, linux-arm64) now build the released `yo-seed` with the **previous release's seed** instead of the TypeScript compiler:

- `install-seed` (env.SEED_VERSION) → `yo compile yo-self/main.yo --release --allocator mimalloc --std-path ./std -o yo-seed`, **seed-era flags only** (the rule the musl job established in #131 — no `--emit-c-to`, `--std-path` pins the checkout's std).
- Their portable-C arms are seed-emitted via the `-o <base>` sidecar form.
- **Pre-release trust-chain gate**: the candidate must re-emit itself (stage-2, emit-only): rc=0, a 100 MB size floor, and a line-anchored `// Failed to transpile` comment check — the hollow-emit trap. Verified against a clean 143 MB self-emit: the pattern matches 0 lines (the phrase appears only inside emitted string literals, never at line start).

macOS legs and Windows stay TS-built for now (recorded user decision): the self-emit peaks at ~12.2 GB after env-sharing (#133), which fits 16 GB Linux runners but not 7 GB macOS runners. Per-module attribution (this PR retires `issues/yo-self-emit-perf-drift-since-r16.md` with the full 2×2 matrix) shows the remaining footprint is emission-phase specialization structures, so the macOS path is step-24 option A (cross-emit fan-out) or C (larger runners). Windows (16 GB) flips a release after Linux proves the chain.

## Also in this PR

- `plans/P2_5_RETIRE_EXECUTION.md`: step-24 second-slice status; step 20 recorded DONE (the ruleset shard contexts were already removed from ruleset 13548862 and the TS shard job already deleted — verified via `gh api` today).
- `issues/retired/yo-self-emit-perf-drift-since-r16.md`: the drift attribution — **no compiler regression** (today's compiler is 1.6-1.7× *faster* than r16's on identical input; the tree got 2.7× heavier at +14% lines) + the per-module `check` table showing ~85% of the emit footprint is emission-phase.

## Verify

- YAML validated; the workflow only changes `seed-bundles` steps behind `matrix.seed_built` conditions — TS legs' steps are byte-identical behind `if: ${{ !matrix.seed_built }}`.
- Real verification is the next `workflow_dispatch` release: both Linux legs green with seed-built bundles, smoke-tested from outside the checkout, stage-2 gate passing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -278,16 +278,29 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # seed_built: true — P2.5 step 24: this leg's bundle is built by the
+          # PREVIOUS release's seed (trust chain: previous-release-builds-next;
+          # escape hatch: bootstrap from an older tag, or the src-attic-final
+          # tag while it exists). Linux first: the seed's self-emit peaks at
+          # ~12.2 GB after env-sharing (PR #133) — it fits the 16 GB Linux
+          # runners but NOT the 7 GB macOS runners, so the macOS legs stay
+          # TS-built until P2_5_RETIRE_EXECUTION.md's step-24 option A
+          # (cross-emit fan-out) or C (larger runners) is picked. Windows
+          # runners have 16 GB, but that leg keeps the proven TS path one
+          # release longer (the windows porting tail is the youngest).
           - os: ubuntu-latest
             target: linux-x64
             experimental: false
+            seed_built: true
           # linux-arm64 promoted: green on its first v0.2.0 run.
           - os: ubuntu-24.04-arm
             target: linux-arm64
             experimental: false
+            seed_built: true
           - os: macos-latest
             target: macos-arm64
             experimental: false
+            seed_built: false
           # macos-x64 runs on macos-26-intel — the LAST Intel macOS GitHub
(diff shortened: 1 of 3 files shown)


shd101wyy:
Second commit adds **option A** (recorded user decision): the macOS legs are now cross-emitted.

- New `seed-cross-emit` job (ubuntu, previous seed, 32 GB swap): emits each macOS target's bundle C (mimalloc) as an artifact plus its libc portable arm directly under `portable-c-<target>`.
- New `seed-bundles-macos` job: downloads the C, clang-compiles it natively with the compiler's own macOS flag set, assembles, and runs the same outside-the-checkout smoke as every leg.
- **Purity verified locally**: with the allocator held constant, an explicit `--target aarch64-macos` emit is byte-identical to the native default emit on an arm64 Mac — the emit is a pure function of the triple.
- `publish-release` now needs `seed-bundles-macos`; `portable-c` needs `seed-cross-emit`.

Remaining TS in seed-bundles after this: the **windows leg only** (16 GB runners — native seed build one release after Linux proves the chain), plus the Pages bun island (B7).

PR #137, opened 2026-08-17
P2.5 step 24 (final leg): windows-x64 goes self-hosted via cross-emit; SEED_VERSION → v0.2.9

shd101wyy:
**DRAFT — merge-blocked until the v0.2.8 release publishes** (install-seed fails hard on a missing bundle, by design).

The last TS-built bundle leg flips to the previous-release seed (16 GB Windows runners hold the ~12.2 GB self-emit); the stage-2 gate becomes exe-aware; `SEED_VERSION` bumps to v0.2.8 — the first release whose Linux legs were themselves seed-built and whose macOS legs were cross-emitted.

After this merges, bun/node in release.yml survives only in the Pages step (the documented B7 island) — which makes **Group E (src/ deletion) unblocked** on the release.yml side.


shd101wyy:
**Approach changed** (force-pushed): windows-x64 now goes through **cross-emit**, not a native seed build.

The previous version flipped `windows-x64` to `seed_built: true` — a native self-build on the Windows runner. That runs **three** self-compile phases (candidate build, stage-2 re-emit gate, portable-c arm) at the ~12.2 GB this file already documents, and the 32 GB swapfile that makes it survivable is `if: runner.os == 'Linux'` — `fallocate`/`mkswap` don't exist on a Windows runner. This PR's own commit message records the failure mode: *"stage-1 build does not fit a swapless 16 GB runner … runner-VM death"*. GitHub's Windows runners are 16 GB, swapless.

The comment claiming "the windows CI job now builds the native compiler non-gating" was **stale** — that build was removed when stage-1 went Linux-only, so nothing had ever proven the native path.

**Note this class of risk is invisible to the checks on this PR** — `release.yml` doesn't run on PRs. A green tick here never meant the leg worked.

So windows-x64 joins the macOS legs on the proven cross-emit route: heavy emit on Linux (where the swap is), Windows only clang-compiles and smoke-tests. `seed-bundles-macos` → `seed-bundles-cross`, since it is no longer macOS-only.

Windows keeps **mimalloc** — #152's system-allocator switch was measured on macOS only, and the allocator is now a per-target matrix field so the emit route can change without silently changing the allocator too. The leg also asserts the PE machine type is `0x8664`: chocolatey's llvm can install a clang whose default target isn't the runner's, and a bundle named `windows-x64` containing the wrong architecture would install and run while emitting for the wrong one.

PR #140, opened 2026-08-17
RC header split: 16 B small header for cycle-incapable types (−0.7 GB, both compilers)

shd101wyy:
## What

Phase 1 of `plans/backlog/RC_HEADER_SPLIT.md` (user-directed): reference-semantics types that can never form RC cycles — exactly the ones whose constructors never `__yo_gc_register` — get a 16-byte RC header instead of 56.

Two findings made this branch-free and registry-free:
1. **Every GC visitor early-returns on `!(gc_flags & __YO_GC_TRACKED)`** — an untracked object's `traverse_fn` and 32 B of GC list-pointers were never read by anything.
2. The big header is **reordered so `dispose_fn` sits directly after the packed count word**, making the small header a strict prefix: `__yo_incr_rc`/`__yo_decr_rc`/borrow checks/visitor flag-reads are layout-agnostic — zero new branches in the hot path.

`typeUsesSmallRcHeader` (both compilers' `types/utils`) drives the per-type choice at the struct/enum member sites and ctor inits. Enum registration stays per-variant; header choice is per-type. Atomic-RC types conservatively keep the big header in phase 1.

## Measured

Self-emit (`compile yo-self/main.yo --emit-c`): **12.21 → 11.53 GB peak (−0.7 GB), 227 s wall**; 317 of ~396 ref types go small (the whole ArrayList army: 80→40 B each).

The plan's −3.31 GB projection was based on a flawed census: its −1 counters lived in dispose *functions*, but inline RC drops free objects without calling the mapped dispose, inflating "live" counts for hot-path types (the tell: "live at exit" exceeded peak footprint, impossible for true liveness). The plan doc records the correction; the follow-up instrument is an allocator-boundary size-class histogram at the high-water mark.

## Verification

- FIXPOINT_HOLDS (stage-2 C compiles, stage-3 byte-identical — this also proves the TS and yo-self implementations emit identical C).
- gates_fast T1 **failures=0**: `check ./yo-self` 248/248, CLI differential 40 PASS / 0 DIFF, goldens clean, fmt idempotent.
- 10 canary test files green under the split compiler (rc, arc, gc, hash_map, string, closure, async_await, algebraic_effects, iterator_combinators, path); smoke binary + ASan clean.
- Full-corpus hollow sweep running; result will be posted before merge.

### DIFF
--- a/plans/backlog/RC_HEADER_SPLIT.md
+++ b/plans/backlog/RC_HEADER_SPLIT.md
@@ -49,10 +49,19 @@ are inflated ~2x by the instrument — compare only tracked numbers.)
 | 64 B class                                 | 18.1M       | 1.08 GB     | 6.1%  |
 | 112 B (Environment) + 80 B (tracked lists) | 15.8M       | 1.39 GB     | 7.8%  |
 
-**Revised lever ranking:** (1) the ≥4 KB buffer pool is the new #1 — 5.11 GB
-that no object-layout work touches; needs call-site attribution
-(`__builtin_return_address` in the shims + atos) to see whether it is HashMap
-tables, ArrayList data arrays, or something else. (2) The 40 B ArrayList army
+**Revised lever ranking:** (1) ~~the ≥4 KB buffer pool~~ **LANDED same day:
+the pool was `_inject_forall_captures`** — ra1 attribution (shims record
+`__builtin_return_address(1)`; arm64 keeps frame pointers) put 3.87 GB /
+167k live copies on that one function; a 150k-call probe measured 99.2%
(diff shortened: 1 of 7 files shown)


shd101wyy:
Full-corpus hollow sweep: **SWEEP_GATE_OK** (all 188 language files under the split binary, zero new regressions, zero stale allowlist entries). Every local gate has now passed: fixpoint byte-identity, gates_fast T1 failures=0, 10 canaries, ASan, corpus ratchet. Merge is held until the v0.2.8 release dispatches (develop-push discipline — a merge would cancel the release-gate test run).

PR #158, opened 2026-08-18
P2.5 Group E: delete the TypeScript compiler (`src/`)

shd101wyy:
**DRAFT — do not merge yet.** Three things must happen first, in order:

1. **#137 merges** (windows-x64 → cross-emit). Until then `release.yml`'s `!matrix.seed_built` steps are LIVE and invoke bun + `out/cjs`, which this PR deletes. I'll rebase and remove those now-dead steps once #137 lands.
2. **A release is cut.** After deletion the only way to build the compiler is with an existing compiler *binary* — the `src-attic-final` tag preserves source, but you can't compile Yo source without a Yo compiler, so the real bootstrap root is a published bundle. v0.2.9 is six commits back, and "a seed built from an older tree miscompiles current source" has actually happened here (`issues/fixed/seed-built-stage1-miscompiles-current-source.md`).
3. **The `src-attic-final` tag is pushed**, pointing at the last develop commit that still contains `src/`. Four surviving files already cite it.

---

## What this deletes — 322 files

| group | count |
|---|---|
| `src/**` | 287 (282 `.ts` + 5 `.yo`), incl. `src/lsp/` (14) |
| root toolchain | 12 — `package.json`, `bun.lock`, `build.js`, `tsconfig.json`, `jest.config.js`, eslint/prettier configs, `transform_expr_id.py`, **both `yo-cli` shims** |
| `scripts/` | 18 — 13 finished codemods, `probe-parser-parity.ts` (the only script importing `src/`), 4 superseded probes/hooks |
| `.husky/`, `perf-repros/` | 5 |

Residue after this commit: **zero** tracked `.ts`/`.tsx` outside `vscode-extension/`, **zero** tracked `.js` outside `vendor/` and `vscode-extension/`. The 9 Python instrumentation scripts, `vendor/`, the shell harnesses and `scripts/install.{sh,ps1}` all stay.

## Not lost, because it was re-homed

The `// Unknown type:` **mid-declaration** grep. That marker is a codegen fallback yo-self shares (`yo-self/codegen/utils/index.yo:1024,1150`), and a line comment landing mid-declaration eats the following declaration and makes the C unbuildable. It lived in the deleted `ts-unit-tests` job; it now runs in `bootstrap-fixpoint` against the already-emitted `/tmp/yo-stage2.c` — no extra compile, and it now watches the **self-hosted** emit. The neighbouring marker check there is line-anchored and structurally cannot catch the dangerous form.

Verified against two real self-hosted emits: markers present, **zero** mid-declaration hits — so it passes today and is a live ratchet, not a broken gate.

## Accepted losses, stated rather than discovered later

- 25 TypeScript test files (~9k lines) and the two-compiler differential. A `tests/internal` failure can no longer be attributed to "self-hosted divergence" vs "the test is wrong". The CLI goldens, fixpoint gates and hollow sweep replace it.
- Five `tests/*.yo` files named a `src/tests/*.test.ts` as their regression guard — those claims are now **past-tense in the source**, naming the retired gate instead of asserting protection that no longer exists.
- prettier/eslint/husky over `*.{md,yml,json}` (`yo fmt` covers `.yo` only), and `bun run dev`'s watch loop, which has no successor.

## Verified

- All **14 required status-check job names unchanged** — a rename blocks every future PR on a check that never runs.
- `rg '\./yo-cli'` outside `plans/`+`issues/` returns only deliberate "how to read older docs" translation notes.
- YAML parses for every workflow.


shd101wyy:
### Post-deletion bootstrap verified locally

The thing this PR really has to prove is that the tree can still build its own compiler with **no TypeScript present anywhere**. Done, on this branch with `src/` absent from disk:

```
$ ls src            → src/ absent
$ <existing yo binary> compile yo-self/main.yo --release --std-path ./std -o /tmp/yo-s2-postdel
BUILD_RC=0          → 8.3 MB binary

$ /tmp/yo-s2-postdel compile hello.yo --release -o hello && ./hello
built from a tree with no TypeScript
```

So the full loop holds: an existing compiler binary builds this tree's compiler, and that compiler compiles and runs a program. That is exactly post-deletion life — and it's also why the recovery root has to be a published **bundle** rather than the `src-attic-final` tag: the tag preserves source, but source needs a compiler binary to become one.

### Also verified on this branch

- **`release.yml` statically validated** (it never runs on a PR, so it gets no CI signal): every `needs:` resolves, and the one undefined-matrix-key site — `matrix.clang_target` on the two macOS rows — is confined to a `runner.os == 'Windows'` step, so it cannot expand to an empty string.
- **Instruction `applyTo` globs span the rename** (`yo-self/**/codegen/**,src/**/codegen/**`), so the C-codegen instructions keep loading both now and after `yo-self/` → `src/`. A single-era glob would have matched zero files and stopped loading silently, with no error.
- **`scripts/cli-diff-test.sh` is golden-only** with no TS arm left, and `NO-GOLDEN` still counts as a failure — an unscored case cannot pass quietly.

shd101wyy:
### Rebased onto post-#137 develop — the last blocking code change is done

#137 moved windows-x64 to cross-emit, which made `release.yml`'s TypeScript-build steps unreachable. Verified before removing them: **no matrix row sets `seed_built: false` any more** (both remaining `seed-bundles` rows are Linux, `seed_built: true`; macOS and Windows are cross-emitted). So these were dead code pointing at files this PR deletes:

- `Setup Bun`, `Setup Node.js`, `Build the TS compiler`, `Build the native yo binary (TS compiler, one last time)`, `Emit this platform's arm of the portable yo.c` — the five `!matrix.seed_built` steps
- the `release` job's own `Setup Bun` + `bun install`
- `Sync root package.json (while it still exists)` — it self-skipped via `-f`, but the file it guarded is now gone for good

Checked that nothing needed went with them: the **seed-side** portable-c emit survives (so `portable-c`'s required arms are still produced), and `actions/setup-node` remains in the `release` job for the VS Code extension's `npm ci` / `vsce`.

### Residue after this commit

```
tracked .ts/.tsx outside vscode-extension/ : 0
tracked files under src/                   : 0
real bun/node/out-cjs refs in .github+scripts : none (comments only)
all 14 required job names                  : present
```

### Still blocked, deliberately

1. **Cut a release** — after this merges, the only way to build the compiler is with an existing compiler *binary*. The `src-attic-final` tag preserves source, and source cannot be compiled without a compiler, so the real bootstrap root is a published bundle.
2. **Push `src-attic-final`** at the last develop commit that still contains `src/`.

Worth knowing for step 1: the current seed (v0.2.9) was cut **8 hours before** the `_inject_forall_captures` memo landed. Measured on this machine (16 GB, same as the runners), compiling this tree: v0.2.9 takes 475 s / 28.4 GB peak footprint, the current tree's own compiler takes 186 s / 11.4 GB. So the next release's bundle will be a much leaner seed, and the raised CI timeouts in #137 become headroom rather than necessity.

PR #171, opened 2026-08-20
P2.5 Group F: rename yo-self/ -> src/, into the name the TypeScript compiler freed

shd101wyy:
The last P2 item — pending since #158 deleted the TypeScript tree and freed the name. **249 files moved, 120 rewritten.**

## What moved, and what deliberately did not

`yo-self` is both a **path** and a **name**. Only the path moved.

- `yo-self/` → `src/` (536 occurrences) and `./yo-self` → `./src` (21), in **live files only**.
- `plans/`, `issues/`, `code-reviews/`, `outdated/` are **untouched** — the deletion manifest lists them as HISTORICAL, and rewriting them would falsify the record. AGENTS.md gets a translation table instead.
- **Bare "yo-self" as a name survives**, including the two **required** CI check names:
  - `Bootstrap fixpoint (yo-self self-compile)`
  - ``Self-hosted `test` subcommand (yo-self tier-1 gates)``

  Renaming those would remove two required status checks and **block every PR in the repository**, fixable only by a ruleset edit from outside the PR. Verified intact.
- The 15 `issues/yo-self-*.md` filenames are `yo-self-`, never `yo-self/`, so the path rule cannot reach them.

## Where the mechanical sweep was wrong

A blind rename produces confident nonsense, so every survivor was inspected:

- **Four tautologies** — "renamed `src/` → `src/`", "if a path MOVES (e.g. src/ -> src/)". Rewritten to say the rename *happened*, naming both sides correctly.
- **Five lines falsified history.** The internal tests moved out of `yo-self/tests` on 2026-08-05, when the directory was still called that. Rewriting to `src/tests` both misdates the move *and* collides with the TypeScript compiler's real `src/tests/`. Restored.
- **Three other `src/tests/fixme.yo` references left alone** — the diff confirms the sweep didn't create them, so they genuinely mean the TS tree.

## AGENTS.md

`src/` now means two different compilers depending on which directory you're reading — the most confusing consequence of this change. The note is rewritten as a table, plus the rule that settles any case:

> **The file extension disambiguates, always.** `src/**.ts` is ALWAYS the retired implementation (~276 live docs still cite it — cleaning those is step 30, the docs sweep); `src/**.yo` is ALWAYS the current one.

## Verified

- both required check names intact; **no `yo-self/` path left in any live file**
- all workflows and composite actions still parse
- **all 66 cross-tree imports in `tests/internal` resolve to real files**
- `build.yo` root is `./src/main.yo`; `.gitignore` entries moved with the tree
- `yo check ./src` running locally, clean so far — CI's GATE 4 covers it too

Draft until the local check finishes and CI reports.


shd101wyy:
Local verification complete: `yo check ./src` → **248/248 files passed, rc=0**. The two `error` matches in the log are filenames (`src/error.yo`, `comptime_expect_error.yo`), not diagnostics.

PR #172, opened 2026-08-20
p3: add a stack-sizing probe that can actually fail

shd101wyy:
Stacked on #171 (base is the Group F branch, so the diff shows only this change).

The last unvalidated item on the musl-only checklist. **Two previous attempts were inconclusive** — and correctly said so rather than passing green.

## Why `check ./yo-self` can't answer it

That workload needs **under 1 MB** at `-O2` — LLVM stack coloring shrinks frames ~100× versus the `-O0` case that exposed the original ceiling. A downward sweep over 1/2/4/8 MB returned `0 0 0 0` for *both* libcs. And 8 MB is useless as a probe value regardless, since **8 MB is also what the silent fallback yields** when `pthread_create` fails.

## The trap this encodes, having fallen into it first

"Non-tail recursion" is **not** sufficient. `n + recur(n - 1)` is linearised by LLVM's accumulator tail-call transform, because addition is associative. The probe then cheerfully reports 500,000 frames fitting in 1 MB — **2 bytes per frame**, i.e. no recursion happened at all.

Only the bytes-per-frame arithmetic gives that away. A probe that cannot fail proves nothing. The fix is an `inout` local whose address escapes, pinning one real frame per level.

## Measured (macOS/arm64, v0.2.9 bundle)

| `YO_MAIN_STACK_MB` | rc |
|---|---|
| 1 | 138 (crash) |
| 4 | 138 (crash) |
| 16 / 64 / 256 | 0 |

Threshold between 4 and 16 MB — consistent with 500k frames at ~24 bytes. **The request is honoured there.** Runs in seconds, versus the 76 minutes the A/B cost without answering this.

## What remains

The same probe against the **static musl bundle on Linux** — the actual open question. That belongs in the musl-only migration PR, as the evidence required before dropping the glibc legs.

## Also

Quiets the expected crash. The shell that *waits* on a signalled child prints "Bus error"/"Segmentation fault", so neither redirecting the command nor wrapping it in a subshell suppresses it — the run happens in a child shell that echoes its own rc. Otherwise a CI log shows what looks like a real failure on the deliberately-failing arm.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,82 @@
+# Contributing to Yo
+
+The `Yo` compiler is **self-hosted**: it is written in Yo and lives in
+[`src/`](./src/). Building it needs an already-installed `yo` binary
+(get one from the [install script](./README.md#install-script-recommended)) plus
+a C compiler.
+
+Yo is primarily developed on the Steam Deck LCD (Linux). The compiler currently transpiles Yo to C; to produce
+machine code you must have a C compiler (for example `gcc`, `clang`, `zig`, `emcc`, etc).
+
+Please install [nix](https://nixos.org/download.html) and [direnv](https://direnv.net/) before proceeding.
+
+The dev environment is defined in [shell.nix](./shell.nix). You can also manually install the dependencies listed in the file.
+
+## Setup
+
+```bash
+$ cd Yo
+$ direnv allow . # Run this command to activate the nix shell.
+                 # You only need to run it once.
+```
+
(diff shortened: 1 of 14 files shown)


shd101wyy:
Correction: the stack-probe numbers were measured with the **v0.2.11** bundle, not v0.2.9 as the PR body says — the local install had been upgraded before the probe ran. The result is unchanged (rc=138 at 1/4 MB, rc=0 from 16 MB up), but the attribution matters given the whole point of the seed bump is that those two versions behave differently. Fixed in the plan doc.

PR #181, opened 2026-08-20
chore: bump vendored mimalloc v3.3.2 -> v3.5.1 — Windows x64 + arm64 work again

shd101wyy:
Supersedes this PR's original v3.5.0 bump, rebased onto develop via merge (606 upstream mimalloc commits kept, plus everything develop grew since 2026-08-20). Four commits: the develop merge, the v3.5.1 bump, a Windows seed-build fix found while verifying, and the measured allocator A/B.

## Why v3.5.1 and not v3.5.0 / dev3

The 2026-08-20 investigation concluded upstream had **no source fix**: v3.3.2 failed arm64 (`__ldar64`/`__stlr64` are cl.exe-only intrinsics; clang defines `_M_ARM64` in MSVC-compat mode), and v3.5.0 broke windows-**x64** on top (`internal.h` `page->self` returning `uintptr_t` as `mi_page_t*`). That changed in v3.5.1:

- `7cca1b54` — clang-cl `__ldar`/`__stlr` workarounds (PR #1379): every `_M_ARM64` call site now guards `!defined(__clang__)` with the memory ordering preserved
- `8cc8a2aa` — C-mode builds with MSVC or clang-cl (PR #1380)
- `604c252a` / `0a4ea88d` — clang-on-Windows compile + link fixes (issue #1370)
- the x64 pointer-type breakage fixed as `mi_atomic_load_ptr_acquire(mi_page_t, &page->self)` (`internal.h:800`)

Upstream also grew plain-C (`MI_NO_USE_CXX`) clang-cl CI including ARM, so the exact route Yo uses is exercised upstream now. **dev3 adds nothing beyond v3.5.1** (tip = v3.5.1 + a CMake default + release notes), so v3.5.1 is the pin.

## Windows x64 AND arm64 build and run — no local patches, plain C11

| check | result |
| --- | --- |
| `clang -std=c11 -fsyntax-only` on `static.c`, both Windows triples (clang 21.1.8) | clean |
| seed → stage-1 build, `--allocator mimalloc`, windows-x64 | OK (2 × `"mimalloc: "` literals) |
| stage-1 `check ./src` (relative `--std-path ./std`) | **263/263 files OK** |
| stage-1 → stage-2 self-compile (`--allocator mimalloc`) | OK, linked against mimalloc |
| cross-emit `aarch64-pc-windows-msvc` + `clang -c` of yo C and `static.c` | both objects clean |

## Found while verifying: seed builds were broken on Windows entirely — fixed here

Unmodified develop fails on Windows before mimalloc is ever reached: every seed-gated build dies with a wall of `requires 'pragma(Pragma.AllowMacroDef);'` errors. `_lex_abs_path` in `src/evaluator/memory_safety.yo` only treated a leading `/` as absolute, so Windows absolute paths were cwd-joined into `<cwd>/C:/...` garbage and the std exemptions of the macro-def/overload gates never matched when `--std-path` arrived relative (`./std`) — exactly what the repo's own build invocations pass. CI never noticed because the seed-gated legs are Linux-only. Commit 2 fixes it via `std/path` (POSIX behavior unchanged), adds `tests/internal/memory_safety_paths.test.yo` (**verified failing pre-fix: 9/12; green post-fix: 12/12**), and records everything in `issues/windows-lex-abs-path-voids-std-exemption.md`.

## Allocator A/B (commit 4): system stays — now on Windows numbers

The decision to ship Windows bundles with `system` originally rested on transplanted macOS numbers. With mimalloc buildable again, the A/B ran natively (windows-x64 host, clang 21.1.8, stage-1 twins from the same seed and tree, 3 reps/arm, min-of-reps, `scripts/bootstrap/measure-windows.ps1`):

| workload | mimalloc v3.5.1 | system | wall verdict |
| --- | --- | --- | --- |
| `check ./src` | 455.58 s / 13,391 MB peak | **340.10 s** / 15,447 MB | system +25%, spreads disjoint |
| full-tree emit (CI/release hot path) | 511.88 s / 12.8–16.3 GB peak | **368.74 s** / 18,490 MB | system +28%, spreads disjoint |

**System is decisively faster on wall on both canonical workloads — the macOS pattern reproduces on Windows — so the shipped-bundle decision stays `system`** (re-affirmed in `plans/WINDOWS_ALLOCATOR_DECISION.md` with the tables). Mimalloc's consistent win is peak RSS; that is why it keeps paying its way on Linux, and why it stays available to Windows users via `--allocator mimalloc`, now CI-exercised per-PR on both Windows native legs (`test.yml` `test-native` links `static.c` on the mimalloc-flavored windows emits — the arm64 runner's clang 22 treats the pointer-type class that killed v3.5.0 as a hard error). The sampler came back from the pruned A/B workflow with its peak capture fixed for modern .NET (the post-exit `PeakWorkingSet64` read silently returns empty there).

Docs updated: arm64-intrinsics issue `retired/` → `fixed/` (with the per-version verification matrix), `plans/WINDOWS_ALLOCATOR_DECISION.md` (measured tables, revisit conditions 1+2 closed), stale `release.yml` comment blocks (including the wrong "windows-arm64 CANNOT link mimalloc" assemble note), AGENTS.md + `src/README.md` internal-test counts.

Merged by admin per repo-owner call to save the CI cycle; the fixpoint byte-identity argument still holds (the path fix is evaluator-only and codegen-neutral; the vendor bump does not touch emitted C), and the Windows mimalloc flavor the CI legs will now compile is exactly the object-compile shape verified above on both triples.


shd101wyy:
## Blocked — v3.5.0 breaks the Windows build

Do not merge as-is. Evidence from A/B run 32353054544, which clang-compiled these sources for Windows targets (nothing in `test.yml` does, which is why this PR's own CI is green on Linux/macOS and would have let it through):

**windows-x64**, which builds fine on v3.3.2:
```
vendor/mimalloc/include/mimalloc/internal.h:792:10: error: incompatible integer to
  pointer conversion returning 'uintptr_t' from a function with result type
  'mi_page_t *' [-Wint-conversion]
    return mi_atomic_load_acquire(&page->self);
```
The immediately preceding run linked the same target with the same command under v3.3.2 and succeeded, so this is attributable to the bump alone.

**windows-arm64** gains the same `internal.h:792` errors *on top of* the existing `__ldar64`/`__stlr64` ones.

Since windows-x64 is `experimental: false` in `seed-bundles-cross` and `publish-release` needs that job, merging this would block releases. (It briefly did — the bump leaked into #182 via a submodule that survived a `git checkout -B`; reverted in #183.)

## The useful finding underneath

Route B in that run — compiling `static.c` as **C++** — **eliminated the `__ldar64`/`__stlr64` errors entirely**, leaving only `internal.h:792`. That is direct evidence that upstream's `MI_USE_CXX` route does what it claims for us, and that the remaining blocker is v3.5.0-specific rather than route-specific.

So the ordering is: switch Windows mimalloc compilation to the C++ route **first** (which likely fixes arm64 on v3.3.2 too), and only then revisit this bump. Not "never update" — update after the build route is fixed.

Unless the pending A/B says mimalloc is not worth carrying on Windows at all, in which case `--allocator system` for both Windows targets dissolves this entire problem and the bump becomes uncontroversial.

Converting to draft until one of those resolves.

PR #252, opened 2026-08-24
codegen: spell `unit` as a one-byte placeholder in C storage positions (D3.8 complete)

shd101wyy:
## The bug

C has no zero-sized object type, but the Yo→C type mapping renders `unit` as
`void` unconditionally. That is legal in **return** position and harmless where
the value is erased — but nothing stopped it reaching a **declaration**:

| shape | emitted C |
| --- | --- |
| `fn(x : unit) -> i32` | `int32_t f(void x)` → *argument may not have 'void' type* |
| `struct(a : i32, b : unit)` | `void b;` → *field has incomplete type 'void'* |
| `(i32, unit)` | same |
| generic at `T = unit` | `void` param + `void` local + `f(xs, )` → *expected expression* |

So `ArrayList(unit)`, `HashMap(K, unit)` and `println(())` — whose parameter is
`value : T` — did not compile. Enum payloads (`Option(unit)`, `Result(unit, E)`)
already worked, because the enum emitter counts non-unit fields and skips the
rest.

## The fix

`get_storage_type_string` spells a type whose C rendering is exactly `void` as a
one-byte placeholder, in **storage positions only** — parameters, struct/tuple
fields, cast targets — and the matching empty argument slot is filled with `0`.
Return position keeps `get_type_string`; pointer types spell `void*`, not
`void`, so both are untouched, and a ref/`inout` unit parameter still takes
`void*` with the `((void*)0)` argument from #250.

The guard therefore fires **only where today's output is already invalid C**,
which makes this fixpoint-neutral *by construction*. Verified rather than
assumed: 8 representative corpus files re-emitted after the change are
**byte-identical** to their pre-change sha256.

Field count and order are preserved rather than erasing unit fields, so struct
layouts, both constructor parameter lists and the compound literal stay aligned
— a missed site fails loudly at the C compiler instead of silently
desynchronizing an ABI. (Full erasure was considered and rejected for exactly
that reason: it changes the parameter ABI of every function, closure, dyn
wrapper and async state machine, where a mask divergence at an indirect call is
silent UB rather than a diagnostic.)

## Result

All nine measured shapes compile **and run**, including `println(())`,
`ArrayList(unit)` and `HashMap(String, unit)`. This closes the D3.8 carve-out in
`plans/STD_API_AUDIT.md`.

Red-first regression test: `tests/unit_as_value_type.test.yo` (6 cases, with an
`Option(unit)`/`Result(unit, E)` no-regression witness).

## Deliberately not fixed

`[unit; N]`, unions, dyn vtables, module-level unit globals, async state-machine
slots, and a `unit` parameter of `main` — each fails today exactly as it did
before, each takes the same one-line recipe. Filed with two pre-existing defects
found while mapping (`is_void_type` never matches `.Unit`, so a unit-returning
`dyn` method emits `return` inside a `void` wrapper; and `unit` is zero bits, so
`ArrayList(unit)` mallocs 0 bytes) in `issues/unit-zst-residual-gaps.md`.

## Validation

Local battery in progress at PR time: `check ./src` 262/262, build, repro gate
9/9 (rc=0, ftt=0, run=0), byte-identity 8/8 identical, `tests/fmt.test.yo` 6/6,
`tests/unit_as_value_type.test.yo` 6/6. Full suite + gates + bootstrap fixpoint
running; I'll confirm before merging.

### DIFF
--- a/issues/fixed/unit-typed-params-and-fields-emit-c-void.md
+++ b/issues/fixed/unit-typed-params-and-fields-emit-c-void.md
@@ -1,7 +1,33 @@
 # `unit` in parameter or field position is emitted as C `void`
 
-**Status:** OPEN
+**Status: FIXED 2026-08-24** (branch `codegen/unit-storage-byte`).
 **Found:** 2026-08-24, while completing STD_API_AUDIT D3.8 (`ToString` for `unit`).
+
+## The fix
+
+C has no zero-sized object type, so a type whose C spelling is exactly `void` is
+now spelled as a **one-byte placeholder** in every STORAGE position — parameter,
+struct/tuple field, cast target — via `get_storage_type_string`
+(`src/codegen/utils/index.yo`), and the matching empty argument slot is filled
+with `0`. Return position keeps `get_type_string`, where `void` is legal, and
+pointer types spell `void*` rather than `void`, so both are untouched.
+
+The guard therefore fires **only where today's output is already invalid C**
+(`void x`, `void f(void x)`, `void b;`), which is what makes the change
+fixpoint-neutral by construction: the emitted C for every program that compiles
+today is unchanged. Verified — 8 representative corpus files re-emitted after the
+change, all 8 byte-identical (sha256), and the bootstrap fixpoint holds.
+
+Field COUNT and ORDER are preserved rather than erasing unit fields, so struct
+layouts, both constructor parameter lists, and the compound literal all stay
+aligned; a missed site keeps failing loudly at the C compiler instead of
+silently desynchronizing an ABI.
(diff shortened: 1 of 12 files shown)


shd101wyy:
Local battery complete and green on this branch:

| gate | result |
| --- | --- |
| `yo build` | ok |
| repro gate (9 unit shapes, compile + run) | 9/9 rc=0, ftt=0, run=0 |
| **byte-identity** (8 corpus files re-emitted) | **8/8 identical sha256** |
| full language suite | **2860/2860** |

Golden re-record produced no drift. Merging.

PR #253, opened 2026-08-24
std S1 (chunk 7): collect + rev complete D3.4, with the comptime-type-arg routing fix that unblocked collect

shd101wyy:
Closes the last two items D3.4 owed — `collect` and `rev` — together with the
compiler bug that was blocking `collect`.

## The compiler bug (the actual blocker)

`evaluate_function_call` routed **any** call whose arguments are all `TypeVal`s
into the CTFE type-constructor path. An ordinary function whose only parameter is
`comptime(C) : Type` always matches that, so it never reached the runtime arm —
which owns the only channel that resolves a return type naming a comptime
parameter. Two symptoms, one gate:

```rust
dflt :: (fn(comptime(C) : Type, where(C <: Default)) -> C)((C <: Default).default());
(a : i32) = dflt(i32);
a.to_string()          // Error: No matching call found — the result is under-resolved
                       // (the annotation does not rescue it; with the result
                       //  discarded, codegen says "Failed to transpile")

d6 :: (fn(comptime(C) : Type) -> i32)({ println(`BODY-RAN`); i32(7) });
d6(i32);               // BODY-RAN prints AT COMPILE TIME, then "Expected to
                       // return a compile-time known value"
```

The decisive probe: adding a `dummy : i32` parameter that has **nothing to do
with `C`** makes both work — it only flips the gate.

The gate now consults the callee's **declared** signature (every real type
constructor is declared `-> comptime(Type)`) instead of the shape of the
arguments. Dropping the disjunct outright would not be safe: it is what keeps a
repeat constructor call memoized and stamped with `constructor_func_id`, and
losing that fragments type identity — the era-split class fixed in #247. The
regression test pins both directions, including that `_Box(i32)` keeps one
identity across two instantiations.

## `rev` / `DoubleEndedIterator`

New prelude trait (`Item` + `next_back`), the `IterRev(I)` adaptor — itself
double-ended, so `.rev().rev()` is the identity — and a blanket `rev` in its
**own** impl block, so the bound sits on a bare-SomeT receiver pattern, the only
shape where an impl where-clause is actually enforced during generic-impl
matching. `rev` on a forward-only iterator therefore fails at method resolution
with a clear message instead of late inside specialization.

`next_back` impls:

- all 20 `Range`/`RangeInclusive` forms — the inclusive one mirrors the forward
  canonical-empty sentinel exactly, so `T.MIN ..= x` terminates from the back
  and mixing `next` with `next_back` on one range stays correct;
- `ArrayListIter` and `_ArrayIter`, via a `_back_taken` **counter** rather than a
  back cursor: the forward guard keeps reading `_list._length` live (so pushing
  during a forward-only walk still extends it), and nothing has to coerce the
  comptime value-generic `N` into a runtime field;
- `DequeIter` / `DequeIterPtr` with **zero** new fields — `_pos` + `_remaining`
  already bracket both ends.

Adaptor `next_back` (on `IterMap` etc.) is deferred: a nominal receiver pattern
is not discriminated by its where-clause, so such an impl would claim
`DoubleEndedIterator` whether or not its source is one.

## `collect` / `FromIterator`

The trait keeps two **monomorphic** statics (`from_iter_new`, `from_iter_add`)
and the iteration loop lives in the blanket `collect`. That is what removes the
"trait-generic-method prototyping" the deferral named — std declares such a
method exactly once (`CustomAllocator.realloc`) and has never dispatched it.

The element associated type is named `Elem`, **not** `Item`: the assoc-type
registry is keyed by (type id, label) with no trait discrimination and takes the
first match, so an `Item` here would collide with the collection's own
`IntoIterator.Item` — benign for `ArrayList`, genuinely divergent for `HashMap`,
whose `Item` is `Bucket(K, V)`. `from_iter_add` returns the accumulator so one
signature covers reference collections and the value newtype `String`.

```rust
squares := (i32(1) ..= i32(4)).map((x) => (x * x)).collect(ArrayList(i32));
uniq    := xs.into_iter().collect(HashSet(i32));
joined  := parts.into_iter().collect(String);
```

Impls: `ArrayList`, `HashSet`, `Deque`, `String`.

## Honest limit

`.map(f).collect(C)` at **two different Item types in one module** still trips
the pre-existing shared-stamp pollution
(`issues/iterator-chain-shared-stamp-cross-item-pollution.md`), and no spelling
of `collect` fixes it — the pollution lives in the shared `IterMap` stamp, not in
the consumer. Every mapped chain in the tests stays at one Item type, and this is
recorded in the plan rather than papered over.

## Also

`DoubleEndedIterator` carries no `where(Self <: Iterator(Item := Self.Item))`:
a trait cannot bind another trait's associated type to its own `Self`-projection
(the constraint value must be a type). Recorded in the syntax cheatsheet.


shd101wyy:

| gate | result |
| --- | --- |
| `yo check ./std` | 154/154 |
| `yo build` | ok |
| `tests/comptime_type_arg_binding.test.yo` (red before) | 4/4 |
| `tests/iterator_combinators.test.yo` | **44/44** |
| full language suite | **2874/2874** |

Golden re-record produced no drift. Merging.

PR #255, opened 2026-08-24
evaluator: canonicalize the method-call return stamp so a self-nesting generic keeps one identity

shd101wyy:
Fixes the bug filed with #253: a generic struct instantiated over **itself** got
two C type identities, so `.rev().rev()` failed at the C compiler.

```
error: initializing '__yo_t19' with an expression of incompatible type '__yo_t11'
```

Both structs are `IterRev(IterRev(Range(i32)))`, structurally identical, under two
different eval struct ids — and the second appeared exactly once in the emitted C,
as that function's return type.

## Root cause

`substitute()` KEEPS the def-era struct `id` and `constructor_func_id` while
rewriting type arguments, so **every** `IterRev` instantiation shares one eval id
and the distinction lives in the rendered type key. Two halves of one call then
disagree:

- The **mint** half (specialization) re-evaluates the declared return-type
  expression under its era-suspect trigger, which routes `IterRev(Self)` through
  the CTFE constructor memo and adopts the **canonical** instance.
- The **caller** half has no such trigger — its gate fires only on leftover
  SomeTs, and the substituted return is already SomeT-free — so the call site
  keeps the substitution copy, which for a self-nesting adaptor holds *its own
  id* in its type arguments. `type_key`'s cycle guard is keyed on the bare struct
  id, so it reads that argument as a back-edge and renders it as the bare id.

Result: `gs_<ctor>_struct_<sid>` at the call site versus
`gs_<ctor>_gs_<ctor>_gs_<range>_i32` at the callee. Codegen keys C types by that
string, so one logical type became two.

Two hypotheses were tested and refuted on the way, and are recorded in the issue
so nobody repeats them: the memo key does **not** miss (arguments compare by id
first, then by constructor fid plus recursive era-equality), and the
`rre`/return-type re-evaluation path is **never entered** for `rev`
(`YO_DEBUG_RRE=1` prints seven lines, none for it).

## The fix

One statement in the method-dispatch arm of `evaluate_function_call`, making it
the twin of the trait-operator arm that #247 already fixed the same way:

```rust
m_out_ty = canonicalize_instantiation_via_ctfe_memo(m_out_ty);
```

Placed **after** the receiver adopt, never before. By the time this arm stamps,
the specialization has already registered the canonical instance in the memo, so
the call site adopts exactly what the callee returns.

Why the existing repairs did not cover it: `adopt_receiver_struct_instance` bails
when the return and receiver share an id — always true for a self-nesting adaptor
— and `canonicalize_instantiation_via_ctfe_memo` had exactly one call site, on
the trait-operator path, which a dot-method call never reaches.

## Why nothing else changes

- **Single `.rev()`** — the lookup either finds the canonical instance, whose type
  key is *identical* to the copy's (at one nesting level no id repeats, so the
  cycle guard never fires), or finds nothing and returns the input untouched.
- **Closure-carrying adaptors** (`map`, `filter`, `filter_map`, `flat_map`,
  `take_while`) — their type arguments include the closure type as a top-level
  bare SomeT, and the canonicalizer early-returns on such an argument. Those
  returns are never touched, so closure-capture identity is preserved.
- **Closure-free adaptors** (`take`, `skip`, `enumerate`, `zip`, `chain`) are
  canonicalized, but only ever onto the instance the callee itself registered —
  i.e. toward agreement, the direction #247 established.

## Tests

`tests/iterator_combinators.test.yo`: double `rev`, **triple** `rev` (the guard
against level collapse), the variable-bound receiver form as well as the chained
one, self-nesting on adaptors other than `rev` (`take` of `take`, `skip` of
`skip`), and a self-nested chain composed with `collect` and with `map` — the last
one keeps the closure-SomeT bail exercised on the same shape.

## A residual I could NOT reproduce, recorded with its negative evidence

The memo's identity rules have two raw-id fast paths whose justifying comment
("a struct id is unique per definition") is false for exactly the shared-id
generics `substitute()` creates, so on paper two *different* instantiations used
as memo arguments should compare equal.

I built the predicted trigger — self-nested chains at three different element
types in one module, plus a mapped one — and **it produces correct results**:

```rust
(i32(0) .. i32(4)).rev().rev().collect(ArrayList(i32))   // [0, 1, 2, 3]
(i64(0) .. i64(4)).rev().rev().collect(ArrayList(i64))   // [0, 1, 2, 3]
(u8(0)  .. u8(3)).rev().rev().collect(ArrayList(u8))     // [0, 1, 2]
(i32(1) ..= i32(3)).rev().rev().map((x) => (x * i32(10))).collect(ArrayList(i32))
                                                          // [10, 20, 30]
```

So it is filed as `issues/ctfe-memo-shared-struct-id-fast-path-smell.md` — an
OPEN *hardening opportunity* carrying its negative evidence and the smallest
sound hardening, explicitly **not** a live bug. I would rather record what was
tried than leave a theory in the tree that reads like a defect.

### DIFF
--- a/issues/ctfe-memo-shared-struct-id-fast-path-smell.md
+++ b/issues/ctfe-memo-shared-struct-id-fast-path-smell.md
@@ -1,0 +1,68 @@
+# The CTFE memo's raw-id fast paths rest on a false invariant (NOT reproduced)
+
+**Status:** OPEN as a hardening opportunity — **no reproducer found**. Filed so
+the next person does not re-derive the theory, and does not chase it as a live
+bug either.
+**Found:** 2026-08-25, while fixing
+issues/fixed/nested-same-adaptor-instantiation-identity-split.md.
+
+## The false invariant
+
+Two places in `src/evaluator/calls/comptime_fn.yo` decide two types are the same
+when their struct ids match, without comparing type arguments:
+
+- `_ctfe_types_era_equal` — `if((a_sid.len() > 0) && (a_sid == b_sid), return(true))`
+- `_ctfe_args_equal` — `((a_struct_id.len() > 0) && (a_struct_id == b_struct_id)) => true`
+
+The justifying idea is that a struct id is unique per definition. That is **false**
+for generic instantiations: `substitute()` (`src/types/substitution.yo`) keeps the
+def-era `id` and `constructor_func_id` while rewriting type arguments, so every
+instantiation of one constructor shares a single id and the distinction lives only
+in the rendered type key.
+
+On paper, then, two DIFFERENT instantiations used as memo ARGUMENTS should compare
+equal, and a second lookup should adopt the first's instance.
+
+## Why it is NOT filed as a bug: the predicted trigger works
+
+The predicted trigger — two self-nested chains at different element types in one
+module, so both outer lookups key on an argument carrying the shared id — was
+built and run. It produces correct results:
+
+```rust
+a := (i32(0) .. i32(4)).rev().rev().collect(ArrayList(i32));   // [0, 1, 2, 3]
+b := (i64(0) .. i64(4)).rev().rev().collect(ArrayList(i64));   // [0, 1, 2, 3]
+c := (u8(0)  .. u8(3)).rev().rev().collect(ArrayList(u8));     // [0, 1, 2]
+d := (i32(1) ..= i32(3)).rev().rev().map((x) => (x * i32(10))) // [10, 20, 30]
+       .collect(ArrayList(i32));
+```
+
+`count()` variants of the same shapes are also correct. So either the lookups do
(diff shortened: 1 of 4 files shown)


shd101wyy:

| gate | result |
| --- | --- |
| repro (double rev) | compiles, prints source order |
| single `rev` / `collect` | unchanged |
| full language suite | **2880/2880** |
| corpus goldens | **PASS 155 / GOLDEN-DIFF 0** — no emitted C changed |
| CLI scorecard | 51 PASS / 0 GOLDEN-DIFF |
| full-corpus hollow sweep | **SWEEP_GATE_OK**, 0 allowlist drift |

The zero corpus-golden diff is the key evidence: the canonicalization is a no-op everywhere except the self-nesting shapes it targets. Merging.

PR #256, opened 2026-08-24
std: delete four dead modules + unexport __MutexUnlocker (S2 deletions, round 1)

shd101wyy:
First round of §6 from `plans/STD_API_AUDIT.md`. Every target re-verified by grep
before removal, and one row of the audit table corrected where it overstated how
mechanical the work is.

## Deleted (531 lines)

| Module | Why |
| --- | --- |
| `std/alg/hash.yo` | zero importers, and its docstring's "Used internally by `HashMap` and `HashSet`" was **false**. The FNV consolidation it hints at belongs to the D3.9 Hasher redesign, which rewrites every `Hash` impl anyway — wiring it in now would be thrown away. `std/alg/` is now empty and gone. |
| `std/encoding/punycode.yo` | zero importers **including `url`**; 343 untested, spec-sensitive lines. IDN support should return with tests and a real `Url` integration rather than sit dormant in a std heading for stability. |
| `std/fmt/display.yo` | zero call sites, not re-exported, and a parameterized trait that misuses its `T` as the self type. Also settled against by O5's single-`ToString` decision. **`std/fmt/writer.yo` stays** — `std/net/addr.yo` uses it. |
| `std/collections/list_view.yo` + its test | no `Index`, no iteration, no consumers; superseded by the copying range forms. The `ListView` recommendation in `.github/instructions/yo-syntax.instructions.md` is updated to say so. |

No CLI golden and no non-archived doc referenced any of them (archived plans are
historical records and are left alone).

## Two places where I did NOT follow the table

**`std/collections/linked_list.yo` is kept.** It is the load-bearing half of the
#249 regression trigger — `tests/where_clause_fn_inference.test.yo` instantiates
one where-bound generic at `LinkedList` and then at `BTreeMap`, the minimal
era-copy/GC-trace reproducer. Deleting it forces substituting a collection pair
that **cannot be verified to still reproduce the original bug**, so the deletion
would trade a real regression test for 514 lines. Revisit if that test is ever
re-expressed without it.

**The export-hygiene row is only partly actionable.** `__MutexUnlocker` comes off
cleanly (used only inside `std/sync/mutex.yo`). The rest do not:

- ~90 of the ~101 underscore-prefixed exports are **real C symbol names**
  (`std/libc/*`'s `_exit`, `_Exit`, `_IOFBF`, `_putenv_s`,
  `_NSGetExecutablePath`, and `std/sys/externs.yo`'s `__yo_async_*` shims). They
  must keep them.
- `std/fs/types.yo`'s three converters and `std/encoding/html_entities.yo`'s two
  builders are consumed by a **sibling** module, so unexporting breaks the
  import. They need module-private visibility (which Yo does not have) or a
  rename admitting they are internal-but-shared — a D8 layout question.
- `__YO_THREAD_SYNC_TYPE` is consumed by `std/sync/cond.yo` **and by four
  compiler files**, so it is D7 work.

Both notes are recorded in the plan so the next pass does not re-derive them.

## Validation

`yo fmt --check` clean. Local battery queued behind the #255 battery on this
machine; I will confirm before merging.


shd101wyy:

| gate | result |
| --- | --- |
| `yo check ./std` | **150/150** (was 154 — the four deleted modules) |
| `yo build` | ok |
| full language suite | **2870/2870** |

Merging.

PR #258, opened 2026-08-25
evaluator: keep `inout` ref-ness when a specialized generic's argument folds

shd101wyy:
A **silent miscompile** with an ordinary trigger: a generic function with an
`inout` parameter, called with a compile-time-known argument.

```rust
show :: (fn(generic(T : Type), inout(x) : Option(T), where(T <: ToString)) -> String)(
  match(x, .Some(v) => `Some(${v})`, .None => `None`)
);

show(Option(i32).Some(i32(4)));            // ❌ the argument folds
o := Option(i32).Some(i32(4)); show(o);    // ✅ a named local binds UnknownVal
show(Option(i32).Some(runtime(i32(4))));   // ✅ not folded
```

No trait, method, `self` or temporary is required — I found it through D3.10's
formatter, but the reproducer above has none of that.

## Two manifestations of one lost flag

**1. The body READS the parameter** → `switch ((x).tag)` on a `T*`: invalid C,
caught loudly by clang.

**2. The body FORWARDS it to another `inout` callee** → **silent wrong value**.
`_apply_ref_amp` passes a ref argument by cancelling an existing `(*x)`
rendering; with `is_ref` lost there is no `(*` to cancel, so it takes the address
again and passes `T**` where `T*` is expected. The callee reads a stack address
as a tag, matches no case, falls through a `switch` with no `default:`, and
returns an **uninitialised** value.

That is only `-Wincompatible-pointer-types` — and `yo compile` passes `-w`, so
nothing was reported. This is exactly why `Option(i32).Some(i32(4)).format(">10")`
returned ten spaces instead of `   Some(4)`.

## Root cause

Pointer-ness is decided by two independent channels:

| channel | source |
| --- | --- |
| C **signature** | the specialized `Func` type's `meta.param_is_ref` |
| **body**'s `p` vs `(*p)` | `Variable.is_ref` on the binding in the atom's recorded `ExprInfo.env` |

In `create_specialized_function_inline`, a parameter bound to a folded comptime
constant is re-bound to a runtime placeholder with `add_variable_to_env`, which
**hardcodes `is_ref : false`**. That push lands on the same innermost frame, and
`get_variables_from_env` does not collapse shadows while `_var_read_code` takes
the LAST match — so the placeholder shadowed the correct `inout` binding for the
whole specialized body evaluation, while the signature stayed a pointer.

The identical defect was already fixed once in the sibling binder in
`calls/function.yo`, whose comment describes this exact symptom; `helper.yo`'s
re-bind was missed.

## Fix

Restore `is_ref` (and `is_reassignable`, since an `inout` body may assign through
the parameter) on the re-bound Variable in **both** arms of that loop — the
folded-const arm and its closure-parameter twin — reading `param_is_ref` off the
`func_type` already in scope. This mirrors `check_param`'s own idiom in the same
file. Synthesized func types carry an empty `param_is_ref`, so the lookup yields
`.None` → `false` → unchanged behaviour. 22 inserted lines, no deletions.

Deliberately **not** included: the same re-bind also drops
`is_owning_the_rc_value`, a latent sibling for `own(p)` parameters. It touches RC
drop scheduling and needs its own leak test.

## Tests

`tests/specialized_inout_comptime_arg.test.yo` — reads, forwarding to another
`inout` callee, writing through the parameter, plus named-local and `runtime(...)`
controls that must keep working. **Verified red before the fix** with the exact
`member reference type ... is a pointer` errors.

## Follow-up worth doing separately

`yo compile` passes `-w`, which is what made manifestation 2 silent. A pointer-type
mismatch between generated code and its own prototype is never acceptable output;
at minimum `-Wincompatible-pointer-types` should survive `-w`.

### DIFF
--- a/issues/fixed/specialized-inout-param-loses-ref-with-comptime-arg.md
+++ b/issues/fixed/specialized-inout-param-loses-ref-with-comptime-arg.md
@@ -1,0 +1,104 @@
+# A specialized generic's `inout` parameter loses its ref-ness when the argument is comptime-known
+
+**Status: FIXED 2026-08-25** (branch `fix/specialized-inout-comptime-arg`).
+**Found:** 2026-08-25, implementing STD_API_AUDIT D3.10, where it produced a
+SILENT wrong answer.
+**Severity: high.** One of its two manifestations is a silent miscompile, and the
+trigger is ordinary: any generic function with an `inout` parameter, called with a
+compile-time-known argument.
+
+## Trigger
+
+specialized (generic) callee **+** an `inout(p)` parameter **+** an argument that
+folds to a compile-time constant.
+
+```rust
+show :: (fn(generic(T : Type), inout(x) : Option(T), where(T <: ToString)) -> String)(
+  match(x, .Some(v) => `Some(${v})`, .None => `None`)
+);
+
+show(Option(i32).Some(i32(4)));   // ❌ broken — the argument folds
+o := Option(i32).Some(i32(4));
+show(o);                          // ✅ fine — a named local binds UnknownVal
+show(Option(i32).Some(runtime(i32(4))));  // ✅ fine — not folded
+```
+
+Nothing about traits, methods, `self`, or temporaries is required — the minimal
+reproducer above has none of them.
+
+## Two manifestations of the one lost flag
+
+**1. The body READS the parameter → invalid C, caught by clang:**
+
+```c
+static inline ... f(__yo_t9* self) {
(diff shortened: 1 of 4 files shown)


shd101wyy:

| gate | result |
| --- | --- |
| regression test (**red before**) | 4/4 |
| full language suite | **2879/2879** |
| full-corpus hollow sweep | **SWEEP_GATE_OK**, 0 allowlist drift |

Merging.

PR #331, opened 2026-08-28
std/http + async codegen: C33 — Timeout, TooManyRedirects, ResponseTooLarge are real; three async bugs fixed en route

shd101wyy:
## C33 (plans/STD_API_AUDIT.md §2)

`HttpError.{Timeout, TooManyRedirects, ResponseTooLarge}` were declared, documented and formatted but never constructed. Now:

- `FetchOptions` gains `timeout : Option(Duration)`, `max_redirects` (default 10) and `max_response_bytes` (default **64 MiB**, `0` = unlimited) plus `with_timeout` / `with_max_redirects` / `with_max_response_bytes`.
- `fetch_with` follows 3xx `Location` hops (relative and absolute-path resolved against the origin; 303 and POST-301/302 become GET; `max_redirects` 0 returns the 3xx verbatim), throws `ResponseTooLarge` *before* buffering past the ceiling, and races the whole exchange against a spawned `sleep` (`_fetch_with_deadline`), throwing `Timeout`.
- `_read_http_response` is an `io.async` future — its plain-fn blocking await nested the event loop (C37 below).
- Six loopback-server tests in `tests/http/http.test.yo` (redirect chain, TooManyRedirects, max_redirects 0, ResponseTooLarge, defaults, Timeout). 17/17 pass.

**Release-notes item:** the 64 MiB default response ceiling is a behaviour change for anyone pulling larger bodies into memory (`with_max_response_bytes(usize(0))` lifts it).

## Compiler / runtime bugs surfaced and fixed (all verified RED first)

- **C36** dispatch-mode cond: a sibling arm's *second* await was neither awaited nor extracted (its continuation rode `chained_branches` with no binding; the dispatch emitters read `cbd.branches` only), and `var = io.await(...)` re-assignment neither recorded a target nor stored its future. **Plain `http://` fetch has been crashing (NULL stream) since #322 added the TLS arm** — the CI test only did https. Repro: `issues/repros/async-cond-dispatch-skips-chained-sibling-arm.yo`. `issues/fixed/async-cond-dispatch-skips-chained-sibling-arm.md`
- **C40** runtime: `__yo_async_run_ready_tasks` drained until empty, so a task re-enqueuing itself (a loop awaiting `yield`) starved `__yo_io_poll` forever. Drain bounded to the queue length at entry (the async-`main` loop already capped at 100). `issues/fixed/async-yield-loop-starves-io-poll.md`

## Filed, open

[...]

PR #333, opened 2026-08-28
std/http + async codegen: C33 client rework on top of #332 — async reader, yield-driven deadline race; C36 + C40 compiler/runtime fixes

shd101wyy:
Supersedes #331 (rebased onto develop after #332 landed its C33 first; a rebase needs a force-push this session cannot do, hence the new branch).

## What this changes on top of #332

`std/http/client.yo` is replaced with the #331 implementation:

- `_read_http_response` is an **`io.async` future**, not a plain fn with a blocking `io.await` inside — that shape is a nested event loop and deadlocks when the peer is a sibling task (**C37**, `issues/sync-await-in-plain-fn-nests-the-event-loop.md`).
- The deadline is a **yield-driven race** in its own future (`_fetch_with_deadline`: spawn request + spawn `sleep`, await `yield` until one finishes), **not `std/async.timeout` inside the task** — the blocking combinator is the same C37 hazard.
- Redirects: relative/absolute-path `Location` resolved against the origin; 303 and POST-301/302 → GET; `max_redirects` 0 returns the 3xx verbatim. `ResponseTooLarge` is thrown *before* buffering past the ceiling. Fields are `usize` (#332's `http_limits.test.yo` adapted).
- Six **loopback-server** tests in `tests/http/http.test.yo` alongside #332's live-HTTPS `http_limits.test.yo`. The loopback tests are what exposed the compiler bugs below.

## Compiler / runtime fixes (verified RED first)

- **C36** — #332 filed `issues/second-cond-internal-await-result-not-stored.md` (plain `http://` fetch crashes since the TLS arm); fixed here: `_dispatch_branches` unions `chained_branches` into both dispatch switches, `_chain_additional_remaining` carries the depth's binding, `_find_branch_await_target_variable_id` / `generate_remaining_expr_future` accept `=` re-assignment. Repro `issues/repros/async-cond-dispatch-skips-chained-sibling-arm.yo`. Their issue moved to `issues/fixed/`.
- **C40** — `__yo_async_run_ready_tasks` drained until empty, so a task re-enqueuing itself (a `yield` loop) starved `__yo_io_poll`; drain bounded to the queue length at entry. `issues/fixed/async-yield-loop-starves-io-poll.md`

Open, filed: **C37** (above), **C38** while-with-await inside a match arm (`issues/while-await-inside-match-arm-missing-loop-field.md`), **C39** handler closure with Box capture (`issues/exception-handler-closure-with-box-capture-fails-inference.md`).

## Known CI issue under investigation

On #331 the Linux hollow-sweep leg hung in the new `fetch throws Timeout` test (macOS: 10/10 green; the other five loopback tests pass on Linux). The test carries stderr step markers so this run's log shows where the Linux path stalls. Battery (fresh binary from this tree) running locally; results in a comment.


shd101wyy:
Full local battery on this exact tree (fresh binary from the v0.2.18 seed): tests/http/http.test.yo 17/17 + tests/http/http_limits.test.yo 5/5 (live), check std 166/166, check src 262/262, cli-diff PASS 52 / GOLDEN-DIFF 0, gates_fast failures=0, **FIXPOINT_HOLDS**, suite 3266 passed / 0 `incompatible pointer`, hollow sweep **SWEEP_GATE_OK**. The Linux-only Timeout hang was C41 (`__yo_io_poll` never entered the kernel under DEFER_TASKRUN) — fixed in the last commit; the Linux legs of this run are the gate for it.

PR #334, opened 2026-08-28
fix: store + poll a cond arm's second await when arms are differently shaped (plain-HTTP crash)

shd101wyy:
Fixes the crash where a cond/match arm with a **second `io.await`** (when arms are differently shaped) never stored — or even polled — that await. This is why **every plain-HTTP request crashed** (`http` does DNS-await then connect-await and uses the connect result; the client had only ever been live-tested over HTTPS, whose later awaits are unused).

An arm with an extra await beyond the shared dispatch point rides `chained_branches`; the dispatch **suspension** switch and the dispatch/chained **extraction** only modelled the representative arm. So the other arm’s second await future stayed **cold** (`->result` null) and its result was never stored → null slot → SIGSEGV.

Three fixes in `state_machine.yo`, all on a new `collect_branch_await_targets` (per-await targets, not one-per-branch):
1. the suspension dispatch emits a case for each **chained** arm’s depth-th await (polls/starts its future);
2. `_emit_chained_await_store` extracts a chained arm’s depth-th await result before its remaining code;
3. the dispatch extraction resolves the destination **per await depth**, not from the single first-await target.

Verified: plain HTTP works (`http://github.com` → https 200), synthetic repro returns 49/10, `tests/async/cond_multi_await.test.yo` passes. Also unblocks C33’s http→https redirect hops.

Battery: fresh checks 166/166 + 262/262, suite 3261/3261, gates CLIDIFF clean, **FIXPOINT_HOLDS**, sweep **229 GREEN** (up 2 — two corpus files hitting this pattern now pass).


shd101wyy:
Closing as **superseded by develop's own chained-arm dispatch** — evidence gathered 2026-08-30 while resolving the conflict with current develop:

1. **The motivating crash no longer reproduces.** A live `fetch("http://github.com")` compiled by a develop-HEAD binary runs clean (response received, rc=0). The chained-arm dispatch folded into `_dispatch_branches` (issues/fixed/async-cond-dispatch-skips-chained-sibling-arm.md) covers the never-polled/stored second await this PR fixed.
2. **This PR's own test passes on develop unchanged** — `cond_multi_await` 1/1.
3. **The remaining delta regresses develop**: merging this branch into current develop first produced a `duplicate case value` compile error in the dispatch switch (this PR's chained pass + develop's `_dispatch_branches` both emitting the arms' cases), and after reconciling that, `tests/http/http.test.yo` went 21 passed → 5 failed — the store-side extraction (`_emit_chained_await_store`, per-depth targets) predates and double-handles the newer dispatch machinery.

The test file is salvaged as #368 so the behavior stays pinned on the machinery that now provides it. Thanks for the investigation — the issue record (issues/fixed/second-cond-internal-await-result-not-stored.md) was the map for everything that followed.

PR #337, opened 2026-08-28
async runtime: panic when the event loop is driven from inside a task (C37)

shd101wyy:
Closes **C37** (`issues/fixed/sync-await-in-plain-fn-nests-the-event-loop.md`).

An `io.await` in a non-`io.async` function, `JoinHandle.await`, and the `std/async` combinators (`join_all`/`race`/`any`/`timeout`) are all loops over `__yo_async_poll_step`. From inside a task that **nests the event loop**: every task below on the C stack is frozen until the awaited I/O completes — a silent, timing-dependent deadlock (this is what the C33 loopback tests hit). The runtime now counts task resume frames (`__yo_async_task_depth`, maintained in `__yo_async_run_ready_tasks`, the async-`main` loop and `wait_all`) and `__yo_async_poll_step` aborts when driven at depth > 0:

```
panic: a blocking await ran inside an async task: an io.await in a non-io.async function, JoinHandle.await, or a std/async combinator (join_all/race/any/timeout) was called from a spawned or awaited task. That nests the event loop and can deadlock. Make the function an io.async future and await it; to race tasks from inside a task, poll is_finished() and await yield().
```

A task's cold start (inline in `io.spawn`/`io.await`) runs at the caller's depth and is deliberately not counted — nothing below it is a frozen task.

**Measurement:** the whole language suite (3266 tests) passes with the guard armed — nothing in-tree nests the loop now that #333 made `_read_http_response` a future. gates_fast failures=0, FIXPOINT_HOLDS. New cli-case `async-blocking-await-inside-task` pins the panic (`build run`, rc 1).


shd101wyy:
Revision: the guard is now armed by `YO_ASYNC_STRICT=1` — `yo test` sets it for every test child (tests/ enforced; suite passes strict) and the cli-case sets it via `env=`. Default stays relaxed because the compiler's OWN build scheduler nests the loop (the smoke-test abort on the previous head): filed as **C42**, `issues/compiler-build-runner-nests-event-loop.md` — make those helpers futures, then flip the default. Local: `yo init` + `yo build run` clean by default; cli-case PASS; suite/gates/fixpoint running.

shd101wyy:
Local battery on the env-gated revision (fresh binary): `yo init`+`yo build run` smoke clean by default, cli-case PASS (panic under `env=YO_ASYNC_STRICT=1`), suite 3266 passed with the runner arming strict mode, gates_fast failures=0 / cli-diff PASS 53, **FIXPOINT_HOLDS**.

shd101wyy:
Thread-local revision: the depth counter is `__thread` (worker-thread resumes made main's blocking await look nested — the Linux `thread_pool.test.yo` abort). Local: thread_pool 13/13 ×3, tests/async 21/21, thread_safety 4/4, gates_fast failures=0, **FIXPOINT_HOLDS**. The Linux hollow sweep on this head is the gate.

PR #340, opened 2026-08-28
std: D3.9 Hasher redesign — Rust-style Hasher/Hash, SipHash-1-3 std/hash, keyed maps (+ C43, C44 compiler fixes)

shd101wyy:
Lands plans/STD_API_AUDIT.md **D3.9** (user decision 2026-08-24: full Rust-style `Hasher`). Design record: `plans/HASHER_REDESIGN.md`.

## std
- **prelude**: `Hasher` (`write(buf, size)`/`finish` required; `write_u8..u64/usize`, `write_i8..i64/isize` defaulted to native-endian bytes) and `Hash.hash(self, inout(hasher) : H, where(H <: Hasher))`; all 22 primitive impls, `Option`/`Result`/`Box`, `derive(Hash)` rewritten (struct fields in order; enum variant index as `u64` then fields). Floats still do not hash.
- **`std/hash`** (new): `SipHasher13` (keyed, streaming; pinned against an independent C reference over 0..16-byte keyed messages + default-key vectors), `Fnv1aHasher`, `DefaultHasher`, `hash_one`, `hash_one_with_keys`.
- `String`/`ImmString` write bytes + `0xFF` (prefix-free, as Rust); `Duration` writes nanoseconds; `imm/map` hashes via `hash_one`.
- **`HashMap`/`HashSet`** carry `k0/k1`: `new()` uses fixed keys (reproducible iteration — the fixpoint gate compares emitted C byte-for-byte, so a `RandomState` default is off the table), `with_keys(k0, k1)` for per-instance keys (HashDoS opt-in via `std/crypto/random`), resize/clone carry the keys.
- `src`: `module_global_c_suffix` and chunk assignment use `hash_one` (deterministic under fixed keys).
- Docs: DERIVE_TRAITS + MEMORY_SAFETY (en + zh-CN); yo-design instructions gained the hashing rule.

## Compiler bugs fixed (RED-before regression tests)
- **C43** `issues/fixed/trait-default-inout-self-bound-by-value.md` — a trait `?=` default with `inout(self)` was bound BY VALUE in the per-impl materialized body: `self.field` on a pointer (clang error) or `&(self)` — a `T**` — passed to sibling methods (silent wrong value under `-w`). `impl.yo`'s default-fill binder now restores `is_ref`/`is_reassignable` from `param_is_ref`. Test: `tests/trait_default_inout_self.test.yo`.
- **C44** `issues/fixed/open-import-retypes-integer-constants.md` — `open(import(m))` typed every non-function/non-struct export from its VALUE, so `K :: u64(7)` arrived as `i32` while the named import kept `u64`. `open.yo` now takes the declared type from the namespace struct type. Test: `tests/open_import_constants.test.yo`.

## Seed gating
The v0.2.19 seed still has C43 and the compiler's own maps run `SipHasher13`; the first seed-built compiler with the defaults in play hung in `__yo_main_module_init`. `SipHasher13` therefore spells out every `write_*` (also the faster path). Recorded in the plan and `plans/backlog/SEED_VERSION_AUTOMATION.md`.

## Closed by measurement (v0.2.19)
`generic-trait-method-reads-primitive-inout-self-as-pointer` (the D3.9 blocker) and `spawn-closure-generic-captures-erased-to-void-ptr` (D7's `join() -> T` blocker) both pass their reproducers — moved to `issues/fixed/` as FIXED BY EVENTS.

## Gates (S1 built from this tree)
`check ./std` 167/167 · `check ./src` 262/262 · `gates_fast` failures=0 (two intended goldens re-recorded: prelude top-level expr count 1154→1156, LSP completion type ids) · full cli-diff 52 pass / 0 diff · targeted tests green (hash 14, trait_default 3, open_import 3, derive 38, prelude 18, imm_string 44, imm_map 21, duration 16, hash_map 64, hash_set 63) · fixpoint_only running locally, result to follow.


shd101wyy:
Local fixpoint gate on the merged head (S1 rebuilt from `c20af5c1f`): `STAGE2_RC=0 STAGE3_RC=0 **FIXPOINT_HOLDS**`. (A first run with an S1 that predated the develop merge reported BROKEN — the diff was the #337 runtime text + type-id renumbering, i.e. generation skew, not a real break.)

PR #342, opened 2026-08-28
codegen: per-branch loop-ness for merged cond/match await points (C38) — while-with-await inside one arm

shd101wyy:
Fixes **C38** (`issues/fixed/while-await-inside-match-arm-missing-loop-field.md`): a while-with-await inside ONE arm of a match/cond whose other arm awaits once merged into a suspension point carrying the plain arm's flags, so the state struct never declared `while_loop_N_active` while the looping arm's body wrote it — clang error, `yo check` green.

- **suspension_analysis**: the merged representative records `cond_branch_suspension_in_while`, parallel to `cond_branch_suspension_exprs` (nested representatives flatten their own flags).
- **exprs/async.yo**: the state struct declares `while_loop_<index>_active` when ANY branch loops.
- **state_code_gen**: branches that disagree on loop-ness force dispatch mode.
- **state_machine `_emit_await_suspension_dispatch`**: the active-check / `goto after_while_loop_N` continuation is emitted only for the looping branches; a loop enclosing the whole cond (representative flagged from outside its arm) still applies to every branch.
- **std/http/client.yo**: `fetch_with` holds the deadline race in its `.Some(limit)` arm again — the `_fetch_with_deadline` indirection that dodged the bug is gone; the arm is the production pin.

Tests: `tests/async/while_await_in_match_arm.test.yo` (both arm orders; RED on the pre-fix compiler with exactly the clang error). Regression: `tests/async/*` (channel 4, combinators 7, effect_bundle 2, join_handle 5, mutex 3), algebraic_effects 74, async_loop_buffer_await 4, `tests/http/http.test.yo` 17 — all green.

Gates (S1 from this tree): `check ./std` 166/166 · `check ./src` 262/262 · `gates_fast` failures=0 (corpus 156/156, cli-diff 53 pass / 0 diff) · fixpoint running locally, result to follow.


shd101wyy:
Local fixpoint gate (S1 built from this tree): `STAGE2_RC=0 STAGE3_RC=0 **FIXPOINT_HOLDS**`. A follow-up commit on this branch closes C39 (handler-capture diagnostic) — gates will be re-run on the final head.

shd101wyy:
Pushed two more commits: **C39** closed as by-design with the swallowed handler-capture diagnostic now propagated (`tests/cli-cases/check-handler-captures-runtime-local`), and the race combinator test's loser widened to 500ms (the 40ms/5ms tie flaked on macos-latest in #341). gates_fast on this head: corpus 156/156, cli-diff 54 pass / 0 diff, failures=0; fixpoint re-running.

shd101wyy:
Fixpoint on the final head (S1 built from it): `STAGE2_RC=0 STAGE3_RC=0 **FIXPOINT_HOLDS**`.

shd101wyy:
Superseded by the fresh-branch PR (GitHub created no Actions runs for this PR's last two pushes).

PR #343, opened 2026-08-28
types: declared enums are nominal under exact comparison (C46) + std §6 base64 string API deleted

shd101wyy:
## Compiler — C46 (`issues/fixed/structurally-identical-error-enums-in-two-generic-impls-collide.md`)
`are_types_compatible_exact`'s `EnumT` arm compared enums purely structurally, so two declared enums with identical variants were ONE type to the CTFE memo: `Result(unit, BErr)` was served the `Result(unit, AErr)` instance. Narrowed to plain user code (no std patch needed): `AErr`/`BErr` through a generic `Carrier(E)` with an impl, both instantiated in one module — the second construction failed.

**Fix**: under `require_exact`, two enums whose names are both non-empty and differ are different types. The name is stamped by `Name :: enum(...)`, so eras of one declaration still converge structurally and `Option(i32)`/`Option(String)` are still told apart by the field compare. Residual (documented): same-named enums in two modules stay conflated.

**Filed, not fixed here (own battery): C45** — the NON-exact comparison accepts `Result(i32, AErr)` where `Result(i32, BErr)` is expected (equal-name fast accept, no payload recursion; the struct arm has the same fast path) — `issues/lenient-generic-enum-compatibility-by-name.md`.

## std §6
- `base64_encode_string` / `base64_decode_string` deleted: text is bytes in / bytes out (`base64_encode(s.as_bytes())`, `String.from_utf8(base64_decode(s, exn))`); `decode_string`'s `Result(String, String)` was a stringly-typed error (D1) and its whitespace skipping a lenience `base64_decode` never had. Only tests used them; the tests now pin the byte idiom.
- `HashMapError.KeyNotFound` / `HashSetError.ElementNotFound` trim: unblocked by C46 but **seed-gated** (the compiler imports both collections; the v0.2.19 seed still conflates) — recorded in `plans/backlog/SEED_VERSION_AUTOMATION.md`.
- The `__yo_c_macro_*` dead-extern row was stale (names no longer exist).

## Tests / gates
`tests/nominal_enum_identity.test.yo` (RED on the previous compiler), `tests/encoding/base64.test.yo` rewritten section; hash_map 64 / hash_set 63 green. `gates_fast` (S1 from this tree): corpus 156/156, cli-diff 53 pass / 0 diff, failures=0; fixpoint running locally.


shd101wyy:
Local fixpoint (S1 built from this tree): `STAGE2_RC=0 STAGE3_RC=0 **FIXPOINT_HOLDS**`.

PR #346, opened 2026-08-28
std: encoding/csv (RFC 4180 reader/writer, P1) + codegen: deferred-dup function tail evaluated once (C47)

shd101wyy:
## std/encoding/csv (plans/STD_API_AUDIT.md §7 P1)
`csv_parse` / `csv_parse_with` / `csv_parse_strict` → `Result(ArrayList(ArrayList(String)), CsvError)`; `csv_write` / `csv_write_with`. LF or CRLF terminators, quoted fields with `""` escapes and any bytes inside, optional final terminator, fields never trimmed; the writer quotes only when a field holds the delimiter, a quote, CR or LF, so `parse(write(rows))` round-trips. Typed `CsvError` with byte positions (D1) + `ToString`/`Error`; `CsvOptions` (delimiter for TSV/`;`, writer line ending); strict mode rejects uneven records. 7 tests.

## Compiler — C47 (`issues/fixed/deferred-dup-return-regenerates-the-call.md`)
The CSV test helper `_field :: (fn(rows, r, f) -> String)(rows(r)(f))` surfaced it: `handle_func_call_deferred_dup` (exprs/return.yo) declared `T temp = <call>` for the dup and then **regenerated the whole tail expression** for the return — two evaluations of any side-effecting RC-returning tail, and for a chained `Index` tail a re-declared spill temp (`redefinition of __yo_idx_tmp_N`). The path now returns the temp it generated the value into. Test: `tests/deferred_dup_return_single_eval.test.yo` (RED on develop; a counting `Index` impl proves one evaluation per call).

Gates (S1 from this tree): `check ./std` 168/168 · `check ./src` 262/262 · `gates_fast` failures=0 (corpus 156/156, cli-diff 53 pass / 0 diff) · fixpoint to follow.


shd101wyy:
Merged-head gates (S1 built from the merged tree): tests 2/2 + 7/7, `STAGE2_RC=0 STAGE3_RC=0 **FIXPOINT_HOLDS**`. (Pre-merge head: gates_fast failures=0, corpus 156/156, cli-diff 53/53.)

PR #348, opened 2026-08-28
std: fs/watch (P1) + codegen fixes C48/C49 + macOS fs events name their entries

shd101wyy:
## std/fs/watch (plans/STD_API_AUDIT.md §7 P1)
`watch(path, opts, exn) -> Watcher` over `std/sys/events` (inotify / kqueue / ReadDirectoryChangesW). Events are queued by the runtime callback and read with the non-blocking `poll()` or the awaitable, yield-driven `next(io)` (no nested event loop, so it is safe inside tasks). `FsEventKind` Rename/Change, `FsEvent.name` relative to the watched directory, `WatchOptions.recursive`, `IoError` through `exn` on a missing path, `close()` makes a pending `next` resolve `.None`.

## Runtime (macOS)
The directory snapshot diff now reports **each changed entry by name** (created/removed → rename, modified → change); the vnode's own kqueue flags add a nameless event only when the diff saw nothing; a single-file watch reports the file's own name. That is the contract inotify and the Windows backend already had — before, every macOS event carried `""`.

## Compiler bugs the module surfaced (RED-before tests)
- **C48** `issues/fixed/inout-receiver-on-enum-variant-literal-takes-address-of-tag.md` — `E.B.to_string()` emitted `&(__YO_T0_B)` (address of an rvalue tag); a comptime payload-free enum arg now takes the ref-spill temp path. `tests/enum_literal_inout_receiver.test.yo`.
- **C49** `issues/fixed/unsafe-cast-to-rc-type-is-treated-as-owned.md` — the `__yo_as` (`unsafe.cast`) result was an OWNING temp, so `w := unsafe.cast(ud, W)` in a callback freed the caller's object at scope end; both extern-call arms attach a non-owning temp for the cast. `tests/unsafe_cast_rc_borrow.test.yo`.

## Audit records
fs.watch DONE; `assert_eq` family already present; C48/C49 rows; **WaitGroup KEEP** and **Layout/layout_of KEEP** decisions (D7 / §6 / S5).

## Gates (S1 from the pre-merge head)
`check ./std` 168/168 · `check ./src` 262/262 · `gates_fast` failures=0 (corpus 156/156, cli-diff 53/53) · tests: fs/watch 4, sys/fs_event 5, unsafe_cast_rc_borrow 2, enum_literal_inout_receiver 1. Fixpoint on the merged head to follow.


shd101wyy:
Merged-head fixpoint (S1 built from the merged tree): `STAGE2_RC=0 STAGE3_RC=0 **FIXPOINT_HOLDS**`.

PR #379, opened 2026-09-02
plans: lazy top-level bindings — order-independent :: definitions and impl registration (design)

shd101wyy:
Design doc only (`plans/LAZY_TOPLEVEL_BINDINGS.md` + its AGENTS.md row). No code.

Lifts the no-forward-reference rule as a planned campaign: `::` definitions and `impl(...)` registrations become order-independent within a module via **pending bindings** forced on first reference or at module end; function forcing is two-phase (signature, then body) so mutual recursion type-checks; cycles are errors with a chain; statements (`pragma`, `open(import)`, module globals, `export`) stay strictly ordered.

**P0 lands first and independently**: a check-time diagnostic `forward reference to "X" (defined at line N)` replacing today's silent swallow → hollow stub (the `std/fs/file.yo` incident that motivated this).

**Acceptance for the semantic phases**: byte-identical emitted C on today's corpus (all of it is order-correct, so force order == source order), which keeps the fixpoint argument free and guards against the registry-perturbation family we hit this week.

Cross-references the retracted 2026-08-13 shell attempt and why this design does not inherit its registry problem.


shd101wyy:
Folded into the D5/§5 closeout PR (branch s8/audit-tail-d5-http): the docs-only fast path skips every test job, so the ruleset's required contexts never report and a docs-only PR cannot merge without --admin, which is no longer used. The two commits are cherry-picked verbatim.

PR #390, opened 2026-09-02
codegen: an io.async closure body with no ExprInfo is a hard error (vacuous-compile fix)

shd101wyy:
## What

A type error inside an `io.async` closure body used to compile **completely vacuously**: `check` green, `compile` rc=0, binary runs rc=0 executing none of the body. The def-time anon-body trial (`_trial_eval_anon_body`) swallows the error, no await analysis is stamped, and the sync-future fallback synthesizes a `_sync_fut_t` future whose resume runs nothing and silently completes — the C22 `_sync_fut_t` class (handover §0c scoping, executed here).

## Fix

Poison gates at BOTH `io.async` emitters in `src/codegen/exprs/async.yo`:
- `generate_io_async_sync_call` — fully-hollow body (the silent class; both known repro shapes route here, including a body whose error comes *after* an await: await analysis is stamped only on full trial success)
- `generate_async_block` — the FSM path, same invariant

Both call `codegen_fatal` when the closure's body block carries no `ExprInfo`, with a diagnostic at the body's source position (same stance as `codegen_fatal_expr`: a diagnostic the C compiler can skip is not a diagnostic). 46 lines, one file.

## Repro (before → after)

```rust
do_bad :: (fn(io : Io) -> Impl(Future(unit, Io)))(
  io.async((io : Io) => { b := (i32(1) && i32(2)); println(b.to_string()); })
);
main :: (fn(io : Io) -> unit)({ io.await(do_bad(io), io); });
export(main);
```

Before: rc=0 everywhere, prints nothing. After: `yo: error: …:8:25: this `io.async` closure's body was never fully evaluated — … would run nothing and silently complete.`

## Acceptance (gen-2 rebuild: seed v0.2.22 → develop@19e4b3817 + this change)

- both repro shapes (error before/after an await) fail with the diagnostic
- **no false positives**: good twin prints `false`; a VALID DEAD (uncalled) io.async closure still compiles (a valid body's def-time trial succeeds even when uncalled); nested io.async green
- `compile src/main.yo --skip-c-compiler` green (whole compiler + std through the gated codegen)
- `tests/async_await.test.yo` **188/188**; `async_trait_default_await.test.yo` **4/4**
- full CLI battery **PASS 57 / GOLDEN-DIFF 0 / NO-GOLDEN 0** — includes the new golden case `compile-async-body-type-error`, recorded with the gated binary

## Notes

- The FSM gate's fire branch has no reproducer yet (every failing body routes sync); kept as the same invariant, exercised-silent by every valid FSM async in the suite — stated honestly in the issue doc.
- The trait-`?=`-default variant of a bad async body fails NOISILY at the C level (undeclared hollow type), so it is not part of the silent class; recorded as follow-up polish in `issues/async-body-type-error-compiles-vacuously.md`.

Closes the async-SM follow-up section of `issues/fixed/wrong-arity-call-silently-accepted-version-install-broken.md`.


shd101wyy:
The wasm-leg failure was the gate working as intended — it exposed a **latent silently-vacuous shape**, not a false positive.

`tests/async_unit_tail_await.test.yo`'s `outer_unit` used a **bare** closure param: `io.async((e) => e.io.await(inner_unit(io), e.io))`. The def-time trial cannot resolve `e.io.await` there (the effect bundle `E` is only bound at call sites whose expected type carries a concrete effect; `Impl(Future(unit))` does not), so the swallow was always silent — before this PR the future compiled to a **runs-nothing sync stub** and the test's `assert(true, "completed")` could not tell the difference. Measured with `YO_DEBUG_SWALLOW`: `[anon-swallow] Error: No matching call found with arguments: ((e.io).await)(inner_unit(io), (e.io))`.

Follow-up pushed (`0897c642e`): the test's closure params are annotated (`(io2 : Io) => ...`, the form every other async test uses — the body then evaluates at def time), and the bare-`e` class is filed as `issues/io-async-bare-e-closure-body-never-evaluates.md` (open work: bind the bundle before the def-time trial from the enclosing signature).

Re-verified with the gate binary on the new head: the file's 2 tests pass, and the **full fast suite (`yo test ./tests --exclude tests/internal --exclude tests/cli-cases --bail`) is green (rc=0)** — the same battery that failed before the test fix. The generic-fn shapes (`put_all_generic` with its where-clause) were never affected: their closure bodies evaluate fine at def time via where-trait resolution.

PR #394, opened 2026-09-03
fix: .io on a bare Io receiver is the identity projection — bare-(e) io.async bodies evaluate and run

shd101wyy:
## What

Closes `issues/fixed/io-async-bare-e-closure-body-never-evaluates.md` (the class #390's poison gate exposed).

An `io.async` closure with an **unannotated** param and a **bundle-style** body — `io.async((e) => e.io.await(f, e.io))` — whose effect binds to the bare `Io` (Step 6b-nested derives `E := Io` from the receiver) could never evaluate at def time: `e.io` is a *bundle* projection and the `Io` struct has no `io` field. Pre-#390 the swallow was silent (the future compiled to a **runs-nothing** stub — the test's `assert(true)` could not tell); post-#390 the gate rejected it.

## Fix

The `.io` projection is now **total**: on a receiver whose type is the `Io` effect struct, `X.io` is the **identity** —
- evaluator twin (`src/evaluator/exprs/property_access.yo`): the access records the object's info (type/value/name/path) instead of a field-miss,
- codegen twin (`src/codegen/exprs/property_access.yo`): returns the object's C text (same shape as the newtype zero-cost field).

Only `.io` is total — `.exn` on a bare `Io` stays a field-miss (there is no exception capability to project). Identity is keyed on the **io-builtin side table** (`lookup_io_builtin_struct_field`), so a user struct with an `io` field is untouched, and bundle receivers (`{io, exn}`) keep their real field projection.

## Acceptance (fixed-binary)

- the repro compiles **and its body runs**: the awaited inner value `7` flows out and the assert holds
- `.exn` on a bare Io still errors; a user struct's `io` field projects normally (prints its value)
- #390's repro still fails (gate intact); good/dead/nested twins still pass
- `check ./std` **172/172**; self-compile green; **full fast suite green (rc=0)**; CLI battery **PASS 57 / GOLDEN-DIFF 0 / NO-GOLDEN 0**; fmt gate clean
- new regression test: *bare-e closure with e.io.await body runs* (`tests/async_unit_tail_await.test.yo`, 3/3)

Note: an orthogonal pre-existing bug surfaced while building the repro (a **module-level global** referenced inside an async closure body emits `use of undeclared identifier` + a mangled capture member — repro available); not addressed here, happy to file it separately if it's not already tracked.

### DIFF
--- a/issues/fixed/io-async-bare-e-closure-body-never-evaluates.md
+++ b/issues/fixed/io-async-bare-e-closure-body-never-evaluates.md
@@ -1,7 +1,10 @@
 # A bare `(e) =>` io.async closure body never evaluates — it compiled to a runs-nothing future, silently
 
-- **Status**: OPEN (the silent form is now a hard compile error via the async
-  poison gate, #390; making the body actually evaluate is the open work)
+- **Status**: FIXED 2026-09-03 — the `.io` projection is now total: on a
+  bare `Io` receiver it is the IDENTITY (evaluator +
+  codegen `property_access.yo` twins), so the bundle-style body evaluates at
+  def time and the future runs it. Pinned by the "bare-e closure with
+  e.io.await body runs" test in tests/async_unit_tail_await.test.yo.
 - **Found**: 2026-09-02, validating #390's poison gate against the test suite
   (`tests/async_unit_tail_await.test.yo`'s `outer_unit`, wasm + macOS legs).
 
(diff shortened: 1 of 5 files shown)


shd101wyy:
The orthogonal bug is now filed: `issues/module-global-referenced-inside-async-closure-undeclared.md` (pushed to this branch).

PR #396, opened 2026-09-03
fix: a module-level global is not a closure capture — closures reach it directly

shd101wyy:
## What

Closes `issues/fixed/module-global-referenced-inside-async-closure-undeclared.md` (filed yesterday while building #394's repro).

A closure body referencing a **module-level global** — plain inline anon closures AND `io.async` closure bodies alike — failed the C compile with both `use of undeclared identifier 'counter'` and `no member named 'counter_m<hash>' in '<capture struct>'`: `trackVariableUsage` recorded the global as a capture, adding a phantom capture-struct member while some use sites still emitted the raw name. Only the module-level *named fn* worked (no capture needed).

## Fix

One guard in `trackVariableUsage` (`src/evaluator/context.yo`): bindings registered in the **module-global registry** (`is_module_level_global` — the same registry the C-name mangler keys on) are skipped from captures. A module global lowers to a process-global C symbol, so the correct lowering is a direct reference. Cross-module globals never reach this path (they resolve through `mod.name` property access).

## Acceptance (fixed-binary)

- all four shapes compile **and run**, each asserting the mutation landed: named module fn, inline anon closure, annotated `io.async` body, and the filed async repro
- `tests/closure.test.yo` **12/12** (two new arms); `tests/async_await.test.yo` **189/189** (one new arm)
- `check ./std` **172/172**; self-compile green; **full fast suite rc=0**; CLI battery **PASS 57 / GOLDEN-DIFF 0 / NO-GOLDEN 0**; fmt gate clean

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -645,7 +645,15 @@ jobs:
       - name: Run tests
         shell: bash
         run: |
-          "$S1" test ./tests --exclude tests/internal --exclude tests/cli-cases --bail --verbose --c-compiler clang
+          # 2026-09-03: the windows-latest / windows-11-arm images' mingw64
+          # toolchain lost libasan ("cannot find -lasan" on every batch),
+          # so ASan is disabled on Windows until the images ship it again —
+          # macOS/Linux keep the sanitizer (issues/windows-images-lost-libasan.md).
+          case "$RUNNER_OS" in
+            Windows) TEST_SAN_FLAG="--disable-sanitize" ;;
+            *) TEST_SAN_FLAG="" ;;
+          esac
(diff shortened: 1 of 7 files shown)


shd101wyy:
The two Windows failures are an **environmental runner-image regression**, not this diff: both `test (windows-latest)` and `test (windows-11-arm)` failed EVERY batch link with `cannot find -lasan` (mingw64 15.2.0 ld no longer ships the ASan runtime; `--c-compiler clang` resolves through the mingw driver there). Uniform across all test files, identical on `--failed` rerun — and develop's Windows legs (run 33704787725) were green minutes earlier, so the image rolled mid-day.

Follow-up pushed (`f60bf25ee`): the two full-suite invocations pass `--disable-sanitize` on Windows only (`RUNNER_OS` conditional); macOS/Linux keep ASan. Filed as `issues/windows-images-lost-libasan.md` with the restore path.

PR #397, opened 2026-09-03
fix: a match arm publishes its deferred drops so a `&&` RHS temp's drop stays in scope

shd101wyy:
## What

A `&&`/`||` whose right operand creates an RC temp (e.g. `String.from(...)`) inside a **non-begin match arm** lowered the operand into a nested `if`-block, but the temp's deferred drop stayed on the arm value's ExprInfo and was flushed after the arm — outside the temp's C scope:

```
/tmp/x.c:17111:10: error: use of undeclared identifier '_file____priv_temp_16011'
```

The test batch runner wraps every test in a match arm, so any test body with this shape broke its whole batch (surfaced by the string-row test work; the shape itself is reachable by ordinary user code on develop — minimal repro in the issue).

## Fix

`generate_case_body`'s non-begin path now publishes the arm value's own `deferred_drop_expressions` as the live `pending_deferred_drops` (concat over the outer set) — the same push the begin path above it and `generate_function_body` already do — so `_emit_drops_for_conditional_branch` claims the branch temps in-branch, and the site-level flush skips them via the existing `short_circuit_handled_drop_var_names` one-shot guard.

## Verification

- minimal repro (match arm + `&&` + `String.from` RHS, plain fn and effectful main): now compiles and runs
- regression arm added to `tests/short_circuit_str_literal_arg.test.yo` (2/2)
- codegen corpus: **156/156 byte-identical** (no codegen output change for existing code)
- gates battery (incl. evaluator shards, macro_expansion), `check ./std` 172/172, `check ./src` 262/262, fmt self-check — all green

Issue: `issues/fixed/match-arm-and-or-rhs-temp-drop-leaks-arm-scope.md` (committed fixed — the fix lands in the same PR).

### DIFF
--- a/issues/fixed/match-arm-and-or-rhs-temp-drop-leaks-arm-scope.md
+++ b/issues/fixed/match-arm-and-or-rhs-temp-drop-leaks-arm-scope.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-OPEN (fix implemented in this session; move to `issues/fixed/` with the PR).
+FIXED 2026-09-03.
 
 ## The error (verbatim)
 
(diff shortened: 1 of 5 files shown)


shd101wyy:
Round 2 (5bc8235f): the first design published the arm's drops into `pending_deferred_drops`, which the effect-unwind path owns — the claim then double-dropped escape-owned temps (json 'lone high surrogate' ASan UAF, visible in CI's first run). New design: separate `arm_value_deferred_drops` field for the claim only, pending untouched; claimed drops removed in place from the shared arm list. json 56/56 under ASan, string 267/267, corpus 156/156 byte-identical.

shd101wyy:
Closing after three fix rounds — the bug is real and fully mapped, but every local fix design broke something subtler:

1. concat-publish into `pending`: double-dropped escape-owned temps (json lone-surrogate ASan UAF — seen in the first CI run),
2. replace `pending` with the arm list: broke unwind propagation,
3. separate `arm_value_deferred_drops` + in-place/copy claiming: clean on json/string/regex + corpus 156/156, but the STAGE-2 SELF-COMPILE fails — the claim, emitted at `&&` chain close, lands in a different C scope than the temp's declaration when the arm body nests blocks (temp declared at stage2.c:1779890 inside a nested if, its claimed drop at :1780190 outside it).

The general fix needs the emitter to track which C block each temp was declared in (a scope stack alongside `declared_c_var_names`) so drops emit at the declaration's own block end — emitter architecture work, not a local patch. Filed OPEN with the full archaeology: `issues/match-arm-and-or-rhs-temp-drop-leaks-arm-scope.md` (branch `fix/match-arm-and-or-drop-scope` keeps the three attempted designs for reference).

The std string-row work that surfaced it (#398) no longer depends on this — its test arms use the begin-form, which exercises the already-correct begin path.

PR #398, opened 2026-09-03
std: string audit row closure — Unicode case, Pattern for rune+Regex, split_once, strip_*, parse_f64/radix

shd101wyy:
Closes the **string** row of `plans/STD_API_AUDIT.md` (and appends the Regex-as-Pattern closure to the regex row).

## String API

- `to_lowercase`/`to_uppercase` are now **Unicode-correct** (Unicode 15.1 tables via `std/string/unicode`, leaf-ified to `to_lowercase_bytes`/`to_uppercase_bytes` so `string.yo` can import it without a cycle); explicitly-ASCII `to_ascii_lowercase`/`to_ascii_uppercase` added beside them (`É` passes through untouched).
- `Pattern` gained `length_in` (bytes of the match starting exactly at `at`) and impls for **`rune`** (in string.yo) and **`Regex`** (in regex/index.yo via a `_find_from` generalization of `exec`'s scan) — every String pattern method now accepts a single rune or a compiled regex: `s.contains(re)`, `s.split(rune(c))`, `s.replace(re, `#`)`, `s.strip_prefix(re)`, …
- `replace`/`replace_all` are Pattern-generic; a **zero-width empty pattern now has Rust semantics**: `"abc".replace_all("", "-")` == `"-a-b-c-"` (it used to return `self` — wrong by both Rust and JS).
- `split_once` — the first std API returning a tuple: `Option((String; String))`; `strip_prefix`, `strip_suffix`.
- `parse_f64` (validates Rust's `f64::from_str` grammar itself, then `atof` — rejects the leading/trailing whitespace, hex floats and trailing garbage atof silently accepts); `parse_i64_radix`/`parse_u64_radix` (Rust `from_str_radix` incl. overflow checks and exactly-`i64.MIN`).
- **Deleted** (breaking, patch-release policy): `panic_dyn`/`assert_dyn` — subsumed by the ToString-generic `panic`/`assert` (the 4 `tests/internal` callers migrated); the malloc'd `to_c_str` `*u8` — its 2 callers now use `to_cstr().ptr()`, making the NUL-terminated `to_cstr()` ArrayList the one C-interop API.
- `std/assert`'s `panic` binds the pointer through a local: `__yo_panic` comptime-evaluates its message and rejects an `.unwrap()` chain (no ExprInfo).

## Notes

- Regex row: "optional-flags `new`" is satisfied by `Regex.compile(pattern)` — expression-valued `?=` defaults are not in the language (probed).
- En route this surfaced a pre-existing codegen bug (a `&&` with an RC-temp RHS inside a NON-BEGIN match arm leaks the temp's drop out of the arm's C scope — the batch runner hit it on these very tests). Root cause mapped across three fix attempts; the correct fix needs emitter branch-scope tracking, so it is FILED OPEN as `issues/match-arm-and-or-rhs-temp-drop-leaks-arm-scope.md` and the affected test arms use the begin-form (which exercises the already-correct begin path). NOT stacked on #397 (that PR is closed — see its analysis).

## Verification

string 267/267 and regex 186/186 **with the unmodified develop codegen** (v0.2.23 seed), plus export-coverage, imm_string, dyn canary, `check ./std` 172/172, `check ./src` 262/262, gates battery, corpus 156/156 byte-identical, CLI 57/57; all modified files `yo fmt`--check clean.


shd101wyy:
Found and fixed a regression in the previous force-push: the panic_dyn/assert_dyn deletion surgery had also swallowed String's `FromIterator` impl (it sat between the deleted fns and the export list) — `collect(String)` failed (`iterator_combinators` caught it on the WASI leg). Restored; structure-audited the whole file against develop (diff is now exactly the intended additions/removals). iterator_combinators 49/49, string 267/267, regex 186/186, check ./std 172/172 — all with the unmodified develop codegen (v0.2.23 seed).

PR #399, opened 2026-09-03
Error diagnostics P1: structured Diagnostic value, one renderer, single-print edges

shd101wyy:
Phase 1 of plans/ERROR_DIAGNOSTICS_OVERHAUL.md (design + decisions in this PR; see the plan doc for the full audit).

## What changes

- **`src/diagnostics.yo` (new)** — the structured diagnostic value (`Severity`, `Span` with 0-based row/rune-column + span length, `Diagnostic` with the P2 `code` slot already rendering as `error[E0308]`) plus the human renderer (rustc-shaped: severity header, `-->` anchor, width-aligned gutter, **full-token underline**) and the one-line short renderer.
- **`src/error.yo`** — `YoError` rebased onto `ArrayList(Diagnostic)`; dead write-only `ErrorKind`/`is_assertion_error` and never-constructed `YoLexerError` deleted. `format_error_message(s)` keep their (token, message) shape — **1028 call sites across 111 files** mechanically trimmed of the dead trailing args — but the location block is no longer baked into the message string, so every error renders its location **exactly once** (D1).
- **Lexer + parser throw the same structured error** (`format_lexer_error`; `make_parse_error` via `format_error_message`): lexer errors now carry the module path, source line and a 1-based anchor (D3); the local `ParseError`/`LexerError` structs are gone, so there is one prefix rule at the CLI edge instead of `yo: error: error:` (D4).
- **Single-print edges (D2)**: module_manager's handlers stash the rendered diagnostic instead of printing; `check` prints per failed file at its own edge, watch rounds likewise, `compile` rethrows the stash so the wrapper carries the underlying text, and `mm_preload_prelude` surfaces a failing prelude. The `check: error in:` leak into compile/lsp/test output is gone.
- Caret underlines the whole token from its start (D5); batch entries after the first render as `note:` blocks (D6).
- **LSP** `parse_error_text` parses the exact renderer shape — end columns are read off the caret run (exact) instead of guessed by identifier heuristics; the doubled-location dedup is deleted with the doubling.
- `tests/internal/error.test.yo` rewritten as the renderer's golden suite (17 exact-string tests); lexer/parser pins updated; `lsp-handshake` golden re-recorded — it loses the leak + doubled location, and the `publishDiagnostics` frames are **byte-identical** to before.

## Verification

- `yo check ./src` — 264/264, rc=0
- `yo compile src/main.yo --skip-c-compiler` — rc=0
- internal suites: error 17/17, lexer 47/47, parser 50/52 — the 2 failures are **pre-existing on pristine origin/develop** under the local toolchain (documented: `issues/parser-multibyte-spec-tests-leak-under-linux-asan.md`; CI-green on the same commit)
- stage-1 built at CI's exact flag set; full CLI corpus: **54 PASS**, 3 `doc-*` diffs that are local git-version stderr text only (no yo output involved), 1 network skip
- live smoke of eval/parse/lex errors confirms the new block, single print, single prefix

## Follow-ups

P2 (codes + bilingual registry + `yo explain` + `--error-format short|json`) and P3 (did-you-mean, LSP typed channel, ICE wrapper, panic locations) land as stacked PRs on top of this one.

### DIFF
--- a/issues/parser-multibyte-spec-tests-leak-under-linux-asan.md
+++ b/issues/parser-multibyte-spec-tests-leak-under-linux-asan.md
@@ -1,0 +1,59 @@
+# The two multibyte spec-peel parser tests leak under Linux ASan (clang 21) — pre-existing, CI-invisible
+
+**Status: OPEN** (documented 2026-09-03 during the P1 error-diagnostics work —
+**not a regression of it**: reproduced on pristine `origin/develop`
+`060acdc26` with the stock seed `v0.2.23`, before any of that work's edits).
+
+## Reproducer
+
+```bash
+git worktree add /tmp/yo-pristine origin/develop
+nix-shell shell.nix --run 'cd /tmp/yo-pristine && ~/.local/bin/yo test ./tests/internal/parser.test.yo --parallel 1 --test-name-pattern "multibyte" -v'
+```
+
+## What happens
+
+Two tests fail:
+
+- `Peel a format spec off a MULTIBYTE template expression`
+- `A spaced colon-pair stays a colon-pair with multibyte content`
+
+Under the nix `clang` 21.1.2 toolchain (Linux x86_64, Steam Deck), the test
+binary's AddressSanitizer/LeakSanitizer run reports at exit:
+
+```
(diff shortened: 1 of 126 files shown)


wongamdoda7-jpg:
I am intrested

PR #402, opened 2026-09-04
std: fs wrapper leftovers + Path Clone/Hash/Ord (fs & path audit rows)

shd101wyy:
Two audit rows advanced.

## fs — the row's `still:` list was stale; re-measured and closed the real gaps

Most of the row had already landed unrecorded (`copy`, `read_link`, `set_permissions`, `try_exists`, `watch`, `remove_dir_all` — including the "compiler implements it twice" note, which was long obsolete). What was actually missing:

- `File.set_len(length, io)` — truncate/extend via ftruncate (Rust's `File::set_len`).
- `File.from_fd(fd)` — wrap an already-open descriptor (Rust's `from_raw_fd` ownership semantics: the descriptor is closed exactly once, by the wrapped handle).
- `DirEntry.path()` — the scanned directory joined with the entry name (Rust's `DirEntry::path`); `DirEntry` gains a `parent : Path` field that `read_dir` fills.

Still open on the row (kept in the audit table): `OpenOptions` builder, Metadata real `btime`/`permissions()`/stop re-stat, walker lazy option, the `AsPath` trait collapsing the _str/_cstr matrix.

## path — `Clone` / `Hash` / `Ord`

- `Clone`: deep copy (sharing the RC handle would alias segment mutation).
- `Hash`: absolute flag + segment count + segments with terminators — consistent with `Eq`, so **Path is now a legal HashMap key**.
- `Ord`: by the normalized rendering (a total order; no platform-collation promise).

## Verification

fs/file 19/19, fs/dir 15/15, path 70/70, walker 8/8, `check ./std` 172/172, `check ./src` 263/263 — all with the unmodified develop codegen (v0.2.23 seed); fmt --check clean.

Learnings folded into the tests: `.Write` is create-or-OVERWRITE (O_TRUNC) — set_len tests open `.ReadWrite`; raw fd reads go through a malloc'd `*u8` buffer (an ArrayList's length does not advance).


shd101wyy:
Local addendum: tests/internal/fetch.test.yo (the DirEntry constructor I touched) 10/10 as well.

PR #407, opened 2026-09-04
PROBE (do not merge): tail-flush drop pinpoint

shd101wyy:
Debug branch for issues/short-circuit-bare-fn-body-operand-temps-leak.md — re-applies the gated tail flush plus a YO_DEBUG_TAIL_FLUSH print probe. The suite-candidate artifact from this run is the probe binary. DO NOT MERGE.


shd101wyy:
Probe complete: the corruption was the v0.2.23 seed-build lottery (issues/v0.2.23-seed-build-lottery-corrupts-shifted-trees.md), not the tail flush — zero emissions fired before the abort, including on eval-only check. Evidence folded into the residual issue. Closing.

PR #414, opened 2026-09-04
std/string: parse_f64 rejected every 3- or 8-byte number; the integer parsers wrapped on overflow

shd101wyy:
Two **wrong-value** defects in shipped v0.2.24 number parsing. Both were found by the std-audit coverage re-measurement, and both were invisible for the same reason: `tests/string/string_parse.test.yo` had 28 tests, **zero** covering `parse_f64` and **zero** covering overflow. Same shape as C34 (`json_parse` reading `"<html>"` as `0` because its number path had no negative test).

## 1. `parse_f64` answered `.None` for `1.5`, `0.5`, `100`, `1e3`, `2.5`, `0.0`, `-1.5`, `1.234567`

`issues/fixed/parse-f64-rejects-every-3-or-8-byte-number.md`

Under shipped v0.2.24:

```
  1.5      -> NONE          1.25    -> Some(1.25)
  0.5      -> NONE          42      -> Some(42)
  100      -> NONE          3.14159 -> Some(3.14159)
  1e3      -> NONE
  2.5      -> NONE
  0.0      -> NONE
  -1.5     -> NONE
  1.234567 -> NONE
```

`_f64_grammar_ok` checks the `inf`/`nan`/`infinity` **words** before the numeric grammar, and both word arms `return`ed **unconditionally**. `rest` is the length after an optional sign, so any 3-byte remainder that is not `inf`/`nan` answered `false`, and any 8-byte one that is not `infinity` answered `false` — the numeric grammar below was **unreachable** for those two lengths. `1.25` (4 bytes) and `3.14159` (7) take the normal path, which is why it looked like "some floats fail".

`parse_f64` shipped as a **v0.2.24 headline feature**.

Fix: the word arms return only when the word actually matches, otherwise fall through. Falling through is safe for rejection too — `abc` reaches the numeric grammar, finds no digits, and answers `.None` there.

## 2. `parse_i32`/`parse_i64`/`parse_u32`/`parse_u64` wrapped instead of rejecting

`issues/fixed/integer-parsers-wrap-instead-of-rejecting-overflow.md`

```
  9223372036854775808:     i32=None u32=None i64=-9223372036854775808 u64=9223372036854775808
  18446744073709551616:    i32=0    u32=0    i64=0                    u64=0
  99999999999999999999999: i32=None u32=None i64=200376420520689663   u64=200376420520689663
```

Each was a hand-rolled digit loop doing `result = result * 10 + digit` with no pre-multiply check. `parse_i32`/`parse_u32` range-check afterwards, which catches ordinary overflow but **not** an input that wraps the accumulator onto an in-range value: `18446744073709551616` is exactly 2^64, so the accumulator lands on `0`, the range check passes, and both answer `Some(0)` for a 20-digit number.

And the two spellings of one operation disagreed:

```
parse_i64("9223372036854775808")           -> Some(-9223372036854775808)
parse_i64_radix("9223372036854775808", 10) -> None
```

`parse_i64` cannot simply grow a guard on its `i64` accumulator: `i64.MIN`'s magnitude (2^63) is not representable there, and the current code only gets `i64.MIN` right **by accident** because the wrap lands on it. A correct signed parse needs a `u64` magnitude plus the `MIN` case — exactly what `parse_i64_radix` already has via `_radix_magnitude`'s `mag > (u64.MAX - d) / radix` check.

Fix: **delegate**, rather than write a fifth copy of the loop. That removed **306 lines** of duplicated hand-rolled parsing for ~40 of delegation, leaving one overflow-correct path. Semantics otherwise preserved exactly. The four methods moved into the later `impl(String, …)` block (the one already holding `parse_f64`/`parse_i64_radix`/`parse_u64_radix`) so the callee is registered before the caller — a same-module forward reference across `impl` blocks is not legal until `plans/LAZY_TOPLEVEL_BINDINGS.md` lands. `String._is_digit_byte` was orphaned by the rewrite and is removed (`src/lsp/diagnostics.yo` has its own private copy, untouched).

## ⚠️ Breaking change

An out-of-range integer input now answers `.None` where it used to answer `.Some(<wrapped garbage>)`. Callers relying on the wrap were relying on a bug, but the change is observable — needs a release-note line.

## Tests — verified RED first

11 new tests. On the **shipped v0.2.24 binary** 9 of the 11 fail, and the 2 that pass are exactly the two guard tests ("still accepts inf/nan/infinity", "still rejects what Rust rejects") — so the fix adds acceptance and rejects overflow **without loosening any existing rejection**.

## Gates

`yo check ./std` 173/173, `yo check ./src` 264/264, `tests/string` 416, `tests/encoding` 160, `tests/http` 34, `tests/cli` 15 — the four parsers have 45 call sites across `src/` and `std/`, including `std/encoding/toml`, `std/http/wire` and the compiler's own CLI.


shd101wyy:
## Blocked on #415 — root-caused, not a defect in this PR

This PR's CI went broadly red with, in every leg:

```
error[E0401]: Variable "Send" not found.
     --> ./std/prelude.yo:3151:11
3151 | impl(u64, Send());
```

That is **not** a prelude problem and not a problem with the change here. Reproduced locally: a stage-1 built from this branch cannot evaluate the prelude, and bisecting by neutering `MAX : u64(18446744073709551615)` moved the failure from line 3151 to line 4098 (`usize.MAX`) — i.e. the trigger is a **large integer literal**, and the reported `Send` is just the next expression after the one that actually failed.

The mechanism: making the integer parsers overflow-checked puts `_radix_magnitude` into the **compiler's own closure** for the first time (`src/evaluator/calls/numeric_type.yo:134` parses every integer literal through `String.parse_u64`). Its guard is

```rust
if(mag > ((u64.MAX - d) / u64(radix)), { return(.None); });
```

and comptime integers are carried as `i64`, so `u64.MAX` is `-1` and that folds to `mag > 0` — every large literal is rejected.

So the real defect is a **compiler bug**: comptime `u64`/`usize` values above `i64.MAX` are divided, modded, shifted and compared with **signed** semantics.

| expr | comptime | correct |
| --- | --- | --- |
| `u64.MAX / 2` | 0 | 9223372036854775807 |
| `u64.MAX % 10` | 18446744073709551615 | 5 |
| `u64.MAX > 1` | false | true |
| `u64.MAX >> 1` | 18446744073709551615 | 9223372036854775807 |

Fixed in **#415** (`issues/fixed/comptime-u64-div-mod-cmp-shr-are-signed.md`), which also files a second, larger finding it exposed: `comptime_assert` never fires inside a function body, so all 1559 `comptime_assert` calls under `tests/` assert nothing.

**Plan:** land #415, rebase this branch on it, re-run. Leaving this PR open and red in the meantime rather than weakening the fix — the overflow rejection is the correct behaviour and the compiler bug is the thing that needed fixing.

PR #441, opened 2026-09-05
issues: json_parse accepts trailing content after the value

shd101wyy:
Issue doc only — **hold until after v0.2.25**.

`json_parse` never checks that the cursor reached the end of input, so `[1,2][3,4]` parses as `[1,2]`, `1 garbage` as `1`, `[1,2]]]]` as `[1,2]` and `null null` as `null`. RFC 8259 §2: a JSON text is ONE serialized value. `serde_json` rejects all four as `trailing characters`.

Same failure mode C34 was filed for (an HTML error page where JSON was expected) — C34 fixed the *number* grammar, but a response that merely **starts** with a valid value still reads as success.

Found while auditing `std/` for the read-overshoot class fixed in #440. Not the same bug: no bytes are misattributed, they are ignored.

**Filed rather than fixed** because the tightening has a real over-rejection risk: `src/lsp/server.yo` parses every JSON-RPC frame through `json_parse`, and whether it can hand over a buffer containing the next frame must be established before the change. The doc records that as the required first step, plus the whitespace acceptance canary.

### DIFF
--- a/issues/json-parse-accepts-trailing-content-after-the-value.md
+++ b/issues/json-parse-accepts-trailing-content-after-the-value.md
@@ -1,0 +1,77 @@
+# `json_parse` accepts trailing content after the value — `[1,2][3,4]` parses as `[1,2]`
+
+**Status: OPEN.**
+
+**Severity: strictness gap, C34 family.** RFC 8259 §2: *"A JSON text is a
+serialized value."* One value — anything after it is a parse error. `json_parse`
+parses the first value, stops, and never checks that the cursor reached the end
+of the input, so it returns `.Ok` for input that is not JSON.
+
+**Found** 2026-09-06, while auditing `std/` for the read-overshoot class fixed in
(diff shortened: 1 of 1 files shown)


shd101wyy:
Superseded by #458, which is merged.

Rather than land the issue record on its own, #458 **fixes the bug** and moves this document to `issues/fixed/` in the same change:

- `_expect_eof` runs after `_parse_value` in both public entry points — `json_parse_bytes` (so `json_parse`/`json_parse_string` throw) and `json_parse_result` (so it returns `.Err`) — reporting `JsonError.UnexpectedChar(ch, pos)`, the way `serde_json` reports `trailing characters`.
- All five shapes from this PR's reproducer are now rejected, verified **red first** (2 failed / 57 passed → 59 passed).
- An over-rejection canary keeps `  [1,2]  ` parsing, since a strictness fix that broke whitespace tolerance would be worse than the bug it closes.

Thanks — the reproducer in this PR is what made the fix straightforward.

PR #443, opened 2026-09-06
D6 step 2: FIX — Windows accept blocked the event loop (overlapped AcceptEx); http/tls suites un-gated on Windows; bisecting probes

shd101wyy:
Stacked on #442. Un-skips `tests/crypto/tls.test.yo` and `tests/http/http.test.yo` on Windows and asserts `tls_available()` there (#442 measured `true` on both runners, legs completed in 28/37 min).

**Purpose: find which test hangs.** The runner streams a `✓` per test under `--verbose`, and the Windows legs now have a 75-minute budget, so a hang ends in a *failed* job with a downloadable log — the last `✓` names the culprit. Not for merge as-is; the fix (and per-test deadlines) follow from what this run shows.

### DIFF
--- a/issues/d6-schannel-hangs-the-windows-test-legs-for-four-hours.md
+++ b/issues/d6-schannel-hangs-the-windows-test-legs-for-four-hours.md
@@ -120,3 +120,20 @@ Both pass on macOS. `http.test.yo` is re-gated with `SkipWindows` so the leg
 reaches `tests/net` (it runs after `tests/http`); `tls.test.yo` stays un-gated
 because its live handshake passed. Whichever probe hangs on Windows names the
 runtime path to fix in `src/codegen/async/`.
+
+### Round 2 (2026-09-06 14:20 CST)
+
+`test (windows-latest)` on #443 **passed in 29 min** with both round-1 probes
+in place: a listener accepting three sequential connections, and a client
+reconnecting after the server closes, both work on Windows x64 (the arm leg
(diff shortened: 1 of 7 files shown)


shd101wyy:
**Root cause found and fixed in `88e88f595`.** `__yo_async_accept_start` (Windows backend) was a plain **blocking `accept()`** on the event-loop thread wrapped in an already-completed future. A server task that awaits `accept` before its peer on the same loop has connected blocks the whole loop — the redirect test and the round-2 probe ("accept pending in a spawned task before the client connects"), deterministically on both Windows architectures. Every pre-existing TCP test connected first, so nothing saw it; the round-1 probes passed for the same reason. Schannel was never at fault.

Fix: overlapped `AcceptEx` (`WSAIoctl(SIO_GET_EXTENSION_FUNCTION_POINTER)` — `<mswsock.h>` typedefs only, no new import library), `is_accept` overlapped record, completion applies `SO_UPDATE_ACCEPT_CONTEXT` + `GetAcceptExSockaddrs` + IOCP association; blocking `accept()` kept only as the no-extension fallback. Gated with `check ./src`, zig cross-compile for x86_64/aarch64-windows-gnu **and link** with `-lws2_32`, native probe run. `http.test.yo` is un-gated again in the same push; `server.test.yo` stays gated for this run.

Follow-up in the same class: `__yo_async_connect_start` is still a blocking `connect()` (harmless on loopback) — `ConnectEx` next.

PR #461, opened 2026-09-07
std: `iter()` yields POINTERS on ArrayList and OrderedMap (D14)

shd101wyy:
## What

Every other std collection — `HashMap`, `HashSet`, `Deque`, `PriorityQueue`, `BTreeMap`, `LinkedList` — already hands `iter()` a `*(T)` into its own storage, and both cheatsheets document `.iter()` that way:

> `| .iter() | *(T) | Borrow via pointer; yields pointers |`

`ArrayList` and `OrderedMap` were the two holdouts, yielding **values**, so `iter()` meant two different things depending on the receiver. **D14** in `plans/STD_API_STABILIZATION.md` §2.

## Changes

- **`ArrayList.iter -> ArrayListIterPtr(T)`**, yielding `*(T)` into the buffer (`Iterator` + `DoubleEndedIterator`, mirroring `ArrayListIter`). `into_iter` already existed with the by-value shape and is unchanged — the old `iter` body was a **byte-identical copy of `into_iter`'s**, so nothing is lost.
- **`OrderedMap.iter -> OrderedMapIterPtr(K, V)`**, yielding `*(MapEntry(K, V))` into the backing `HashMap`'s bucket array. The by-value iterator becomes `into_iter` via a new `IntoIterator` impl — so **`for(map, ...)` works on an `OrderedMap` for the first time** — and `OrderedMapIter`'s bare `next` becomes a real `Iterator` impl.
- **New `HashMap.get_entry_ptr(key) -> Option(*(MapEntry(K, V)))`**, which is what lets `OrderedMap` point *into* the map rather than rebuild entries by value.

## Why it matters beyond consistency

Elements are now **borrowed rather than dup'd**, so iterating a list of RC values costs no refcount traffic, and writing through the pointer mutates in place. Both are covered by new tests:

```rust
for(list.iter(), ptr => { ptr.* = (ptr.* * i32(10)); });
assert(list(usize(0)) == i32(10));   // the list itself changed
```

The pointers alias the storage, so a reallocating mutation mid-walk invalidates them — documented on each iterator type.

## Blast radius

**No `std/`, `src/` or `vendor/` caller used either method** — every `.iter()` in those trees is on a `HashMap`, which is unchanged. So the only call-site churn is in tests.

Two pre-existing tests named `"ArrayList iter - …"` actually exercised `for(list, ...)` (i.e. `into_iter`); renamed to say so, since D14 makes the old names misleading.

`ordered_map.yo` gains `pragma(Pragma.AllowUnsafe)` because it now names a raw pointer type; the pointers come from `HashMap.get_entry_ptr` and are never dereferenced in that module.

## Verification

- `tests/collections` suite: **416/416**
- `yo check ./src`: **266/266**

## Merge timing

### DIFF
--- a/issues/fmt-pointer-type-paren-verdict-is-context-dependent.md
+++ b/issues/fmt-pointer-type-paren-verdict-is-context-dependent.md
@@ -36,12 +36,28 @@ it away in another. Whatever selects between them is NOT the pointer type, the
 enclosing `Iterator(...)`, the `where` clause, or where `MapEntry` comes from —
 those are the same on both sides.
 
-A reduced probe (a fresh file with the same `impl` shape and a one-line `.None`
-body) reports rc=1, i.e. it behaves like `ordered_map.yo`, NOT like
-`hash_map.yo` — so `hash_map.yo` is the outlier and something in its context is
-suppressing the rewrite. The remaining difference between the two real files is
-the shape of the `next` BODY: `hash_map.yo` opens `(` + newline + `cond(`,
-while `ordered_map.yo` opens `({`.
+Three experiments narrow it to a **file-scope** trigger, not a construct-scope
+one:
+
+1. **A reduced probe fails.** A fresh file containing the same `impl` shape
+   (same `where`, same `Iterator(...)`, `MapEntry` imported from
+   `std/collections/entry.yo`) reports rc=1 — it behaves like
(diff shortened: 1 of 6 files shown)


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #462, opened 2026-09-07
std/async: `timeout` returns Result(T, TimeoutError) (D18, first half)

shd101wyy:
## What

`timeout -> Option(T)` collapsed three outcomes into `.None`:

- the deadline fired,
- the task was cancelled on its own,
- and, when `T` was itself an `Option`, the task legitimately returning `.None`.

A caller that wants to retry a slow task but give up on a cancelled one could not tell which had happened. **D18** in `plans/STD_API_STABILIZATION.md` §2.

`timeout` now returns `Result(T, TimeoutError)`:

```rust
TimeoutError :: enum(
  /// The deadline fired first; `timeout` aborted the handle.
  Elapsed,
  /// The task was already aborted (or unwound) while the deadline was still
  /// unexpired, so it never produced a value.
  Aborted
);
```

with `ToString` + `Error` impls, matching every other std error enum.

## Why two variants and not just `Elapsed`

Rust's `tokio::time::timeout` is `Result<T, Elapsed>` — but it wraps a *future*. Yo's `timeout` takes a **`JoinHandle`**, which adds cancellation as a genuinely separate failure. Collapsing it back into `Elapsed` would rebuild the ambiguity D18 exists to remove.

## Implementation note

The two failures are distinguishable **only inside the poll loop** — once it exits, `handle.await(io)` reports `.None` for a deadline abort and a self-abort alike. So the deadline arm records which arm ended the wait, and that flag picks the variant.

## Blast radius

No `std/` or `src/` caller used `timeout`, so the churn is confined to `tests/async/combinators.test.yo`.

Added **"timeout distinguishes an aborted task from an elapsed deadline"** — it cancels a task well inside a generous deadline and asserts `Aborted`. That case could not be expressed at all under the old signature.

`tests/async/combinators.test.yo`: **8/8 pass**.

## Not in this PR

D18's second half (`Thread(T).spawn` carrying its result, `join() -> T`). `Thread` is non-generic today and `Thread.spawn` is wired into the evaluator's closure-specialization and the codegen/parallelism paths, so it needs its own change rather than being bolted onto this one.

## Merge timing

### DIFF
--- a/issues/thread-spawn-callback-returning-a-zst-emits-void-star-from-void.md
+++ b/issues/thread-spawn-callback-returning-a-zst-emits-void-star-from-void.md
@@ -1,7 +1,8 @@
 # `Thread(T).spawn` cannot take a callback returning a ZST — the spawn path emits `void* tmp = <void expr>`
 
-**Status:** OPEN — the codegen bug is real and unfixed. D18 part 2 **DID land**
-by routing around it; see "The workaround that landed" below.
+**Status:** OPEN, and D18 part 2 is **BLOCKED**. The workaround recorded below
+dodges this bug and then hits a SECOND wall — see "Why the workaround does not
+land either".
 
 Found 2026-09-07 attempting **D18 part 2**
 (`plans/STD_API_STABILIZATION.md` §2: *"`Thread(T).spawn` carries its result and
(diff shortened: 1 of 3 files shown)


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #463, opened 2026-09-07
std/collections: `sort` is STABLE, `sort_unstable` is the heapsort (D17)

shd101wyy:
## What

`ArrayList.sort` was a **heapsort under Rust's stable name**. Rust guarantees `sort` is stable and offers `sort_unstable` for the faster non-stable one; Yo had the fast one wearing the stable name, so any code sorting by a secondary key silently lost the primary order. **D17** in `plans/STD_API_STABILIZATION.md` §2.

- `sort` / `sort_by` → a new stable bottom-up merge sort
- `sort_unstable` / `sort_unstable_by` → the existing heapsort, now correctly named

## Implementation

`_merge_sort_indices` sorts an index permutation (bottom-up, ties take the **left** run — that is what makes it stable), then the elements are permuted in place by cycle-following.

Two things worth flagging for review:

**1. The permutation had to be inverted.** The merge sort yields `order(i)` = the *source* index for destination `i`, but the cycle-following swap consumes *destination* indices. Using them directly produced `3 5 1 1 4` from a 5-element list. Verified on paper (`[A,B,C]` with `order=[2,0,1]` → `[B,C,A]`, not `[C,A,B]`) before fixing, and the inversion into `scratch` is commented in place.

**2. `_merge_sort_indices` is deliberately monomorphic** (`usize` indices), not a reuse of the generic `_heapsort_by`. Reusing the generic at `T=usize` emitted C that passed a `size_t` to a closure expecting the element struct — a **compiler specializer bug**, filed with a reduced reproducer in the second commit (`issues/generic-fn-specialized-at-two-types-hands-the-closure-the-wrong-param-type.md` + `issues/repros/`). The monomorphic version sidesteps it; the bug itself is independent of this PR and still open.

The trigger took three attempts to reduce — the first two reproducers compiled cleanly. It needs an **outer** generic instantiating an **inner** generic at a *different* type, with a closure capturing the outer `T`'s data.

## Verification

6 new tests in `tests/collections/array_list_convenience.test.yo` using a `struct(key, tag)` record so stability is observable (equal keys must keep their original `tag` order). **30/30 pass.**

I also suspected the heapsort swaps were leaking RC handles and checked with a `Dispose`-counter witness — they are not. Recording that so it is not re-investigated.

## Merge timing

Breaking (`sort`'s complexity/ordering guarantees change), so it belongs to the **v0.2.28** window — please hold until v0.2.27 is out.

### DIFF
--- a/issues/generic-fn-specialized-at-two-types-hands-the-closure-the-wrong-param-type.md
+++ b/issues/generic-fn-specialized-at-two-types-hands-the-closure-the-wrong-param-type.md
@@ -1,0 +1,65 @@
+# A generic fn instantiated at two `T` hands one closure the other's parameter type
+
+**Status: OPEN** (found 2026-09-07, while implementing D17's stable sort).
+
+**Severity: emitted C does not compile.** Not silent — clang rejects it — but
+the diagnostic names a generated symbol and a `__yo_tN` type, so the user has
+no way to see which of their calls is at fault.
+
+## Symptom
+
+The trigger is narrower than "two instantiations". Both of these compile
+fine:
+
+- a generic `Impl(Fn(a : T, b : T) -> bool)` helper called at two `T` from
+  `main`;
+- the same helper taking a `*(T)` as well, still called at two `T` from
+  `main`.
+
+What breaks is an **outer generic instantiating an inner generic at a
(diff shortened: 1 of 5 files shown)


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #464, opened 2026-09-07
std: `replace` replaces EVERY occurrence; add `replacen` (D10)

shd101wyy:
## What

`String.replace` and `ImmString.replace` replaced the **first** occurrence only, under the name every Rust, Python, Go and JavaScript user reads as replace-all. Nothing at a call site says which it is, so code that means *"normalize all of these"* compiled, ran, and was silently wrong on any input with two matches. **D10** in `plans/STD_API_STABILIZATION.md` §2.

The tree already contained the proof:

```rust
std/path.yo:64        path_str.replace_all(String.from("\\"), String.from("/"))   // correct
src/main.yo:475       joined.replace(String.from("\\"), String.from("/"))         // same job, wrong
src/expr_info.yo:883  module_path.replace(String.from("\\"), String.from("/"))    // same job, wrong
```

## The new shape

| method | meaning |
|---|---|
| `replace(pat, new)` | **every** occurrence — Rust's `str::replace` |
| `replacen(pat, new, count)` | first `count`; `count == 0` changes nothing |
| `replace_first(pat, new)` | the pre-D10 meaning — **deprecated**, one release |
| `replace_all(pat, new)` | synonym of `replace` — **deprecated**, one release |

All four are thin wrappers over a single `_replace_scan` primitive, so the zero-width-match rule (step one rune past each empty match so the scan advances) and the needle-longer-than-haystack guard are stated once instead of twice.

## `Regex.replace` is deliberately NOT flipped

The Rust regex crate's `Regex::replace` is first-match and `Regex::replace_all` is all — Yo already matched Rust there, so the method **on** a `Regex` keeps first-match semantics. It is `String.replace(re, rep)` (the `Pattern` dispatch) that now replaces all, consistently with `String.replace("lit", rep)`.

## The audit, before flipping

I classified all 61 `.replace(` sites rather than flipping blind. **No site in `std/`, `src/` or `tests/` genuinely wanted first-only** — which is itself the evidence that the name was wrong rather than the callers.

Four compiler sites were **already silently broken** and become correct without being touched:

1. **Windows path normalization stops at the first backslash** (`src/expr_info.yo:883,892`, `src/main.yo:475`) — `C:\proj\src\main.yo` → `C:/proj\src\main.yo`. Windows-only, which is why it survived.
2. **`/./` collapsing stops after the first segment** (`src/expr_info.yo:904`) — `a/./b/./c` → `a/b/./c`.
3. **The "whitespace-stripped" pragma scan strips one space and one tab** (`src/main.yo:2585,2620`). The comment states the intent outright and the code does not do it: `pragma( Pragma.SkipWasm )` with extra spaces did not match, so the skip was ignored and **the test RAN on a wasm target**.

Written up with the reachability of each — including **one latent-but-currently-unreachable** case (`_ti_escape_str` escapes only the first backslash and quote, but every caller passes a Yo identifier, which cannot contain either) and **one that looks wrong but is not** (the decl-key sanitizers collide identically under both semantics) — in `issues/fixed/string-replace-first-only-broke-compiler-callers.md`.

The other 8 sites strip a `file://` prefix or a `__self_shell` marker that occurs once, where first and all agree.

## Verification

- `tests/string/string.test.yo` **269/269**
- `tests/imm_string.test.yo` **44/44**
- `tests/regex/regex.test.yo` **186/186**
- `tests/template_string_specs.test.yo` **6/6**
- `yo check ./std` **173/173**

New tests cover `replacen` at counts 0/1/2/past-the-end, `replace_first` as the old meaning, and the empty-search interleave (`"hello".replace("", "X") == "XhXeXlXlXoX"`, matching Rust).

The rule is recorded in `.github/instructions/yo-design.instructions.md` for future sessions. It is deliberately **not** added to `.github/skills/`, since that tree is bundled into `yo init` and would force a re-record of six cli-case golden trees.

## Merge timing


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #466, opened 2026-09-07
std: FromStr + `s.parse(T)` with real parse errors (D12)

shd101wyy:
## What

Eight `parse_*` methods returned `Option`, collapsing **"empty string"**, **"contains garbage"** and **"does not fit the type"** into a single `.None`. A caller that wants to tell a missing config field from a malformed one from an out-of-range one had nothing to branch on. **D12** in `plans/STD_API_STABILIZATION.md` §2.

```rust
`2147483648`.parse_i32()  // .None — overflow? garbage? no way to know
`2147483648`.parse(i32)   // .Err(ParseIntError.PosOverflow)
```

## The new API

- **`FromStr`** trait (Rust's `core::str::FromStr`) with an associated `Err`, so each type reports its own error type.
- **`s.parse(T)`** forwards to `T.from_str(s)`. `T` is written out because it can only be inferred from the *result*, which Yo does not do — so `s.parse(i32)`, not a turbofish.
- **`ParseIntError`** (`Empty`, `InvalidDigit`, `PosOverflow`, `NegOverflow`, `InvalidRadix`), **`ParseFloatError`** (`Empty`, `Invalid`), **`ParseBoolError`**.
- Impls for `i32`, `i64`, `u32`, `u64`, `f64`, `bool` — the six types that had a `parse_*`. The remaining widths are a follow-up, called out rather than silently omitted.
- **`parse_i64_radix_res` / `parse_u64_radix_res`** are the `Result`-returning radix parsers. Rust *panics* on a radix outside 2..=36; Yo reports `InvalidRadix`, because in practice the radix is often user input too.

## Nothing breaks

The eight `Option` methods stay for one release as **deprecated wrappers over the same cores**, so no call site changes and there is exactly one parsing implementation rather than two that can drift. `_radix_magnitude` lost its last caller and is deleted.

## Layering

`string` < `fmt` < `error`, and the pieces had to land at the lowest layer that can name what they need:

| piece | module | why |
|---|---|---|
| enums + `FromStr` + `parse` | `std/string` | `std/error` imports `./string`, so string cannot import back |
| `ToString` impls | `std/fmt/to_string.yo` | first layer that can name `ToString` |
| `Error()` impls | `std/error.yo` | same split `impl(String, Error())` already uses |

## One subtlety, with a regression test

The magnitude scanner overflows `u64` and can only report `PosOverflow`, so the **signed core re-signs it** — otherwise `-99999999999999999999999` would come back as `PosOverflow`. There is a test named for exactly that.

## Verification

- `tests/string/string.test.yo` **275/275** (8 new tests)
- `yo check ./std` **173/173**, `yo check ./src` **266/266**, full `yo build` green — this touches `std/error.yo` and `std/fmt`, which the compiler itself uses

New tests cover empty vs garbage vs positive vs negative overflow, the `i32`/`i64` boundary values (including `i64.MIN`, which *is* representable), unsigned rejecting a leading `-` as `InvalidDigit` (Rust's behaviour), the bad radix, and the deprecated methods still agreeing.

## Merge timing

Additive — nothing breaks — but it belongs with the rest of the D-batch in the **v0.2.28** window unless you would rather take it into v0.2.27.

### DIFF
--- a/std/error.yo
+++ b/std/error.yo
@@ -3,7 +3,7 @@
 // re-exported ALL of string/fmt through std/error — every error-module
 // importer gained println, String, Format, ... — for the two names actually
 // used here).
-{ String } :: import("./string");
+{ String, ParseIntError, ParseFloatError, ParseBoolError } :: import("./string");
 { ToString } :: import("./fmt/to_string.yo");
 /// Standard error trait for typed error propagation.
 /// All error types should implement this trait along with `ToString`.
@@ -17,6 +17,11 @@ AnyError :: Dyn(Error);
 export(AnyError);
 /// Allows using `String` directly as an error type.
 impl(String, Error());
+/// The D12 parse errors: defined in `std/string`, rendered in `std/fmt`, and
+/// made propagable here — each at the lowest layer that can name what it needs.
+impl(ParseIntError, Error());
+impl(ParseFloatError, Error());
+impl(ParseBoolError, Error());
 /// Non-resumable exception handling effect record.
 /// Use `throw` to raise an `AnyError` and abort the current computation.
 /// `throw` is `ctl(...)` — handler body may contain `unwind`, and the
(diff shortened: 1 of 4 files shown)


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #467, opened 2026-09-07
std/encoding: hex, base64 and utf16 decoders return Result (D13, part 1)

shd101wyy:
## What

`hex_decode`, `base64_decode`, `base64_decode_url` and `utf16_to_utf8` are **pure transforms** — bytes in, bytes out — that reported failure by throwing through `Exception`. **D13** in `plans/STD_API_STABILIZATION.md` §2.

That costs twice over: every caller must be inside an effect scope for what is just a parse, and the *reason* for the failure becomes nearly unreachable, because a handler only sees a `dyn` error. The tests show the consequence — they were reduced to asserting proxies:

```rust
r = base64_decode(String.from("Zm9v YmFy"), exn);
threw = (r.len() == usize(0));
assert(threw, "the decode must have unwound before returning bytes");
```

## The new shape

```rust
hex_decode(s)                            -> Result(ArrayList(u8), EncodingError)
base64_decode(s) / base64_decode_url(s)  -> Result(ArrayList(u8), EncodingError)
utf16_to_utf8(data)                      -> Result(String, EncodingError)
```

with `hex_decode_exn`, `base64_decode_exn`, `base64_decode_url_exn` and `utf16_to_utf8_exn` kept for callers already inside an effect scope — each still covered by a test.

**No `std/` or `src/` code called any of them**, so all call-site churn is in tests.

## Asserting on the error found a test passing for the wrong reason

Converting that whitespace test from the proxy to a direct assertion immediately failed — and it was the *test* that was wrong, not the decoder:

`"Zm9v YmFy"` is **9 symbols**, so the `1 mod 4` **length** check fires before the scan ever reaches the space. The old test could not tell, because "it unwound" is all it ever checked. It now asserts `InvalidLength` for that input **and** `InvalidChar(0x20)` for `"Zm9v YmFyIQ=="` (11 symbols, so the scan does run), documenting the precedence between the two checks.

## Verification

- `tests/encoding/hex.test.yo` **9/9**
- `tests/encoding/base64.test.yo` **20/20**
- `tests/encoding/utf16.test.yo` **14/14**
- `yo check ./std` **173/173**

Orphaned `exn` locals left behind by the conversion are removed (18 of them); pre-existing unused imports are left alone.

## Not in this PR — D13 part 2

- **`Url.parse`** is a long function with `exn.throw` threaded through its whole body; converting it means restructuring the parser, not wrapping it.
- **`json.yo` ships two complete parsers** — `json_parse`/`json_parse_bytes`/`json_parse_string` (exn) and `json_parse_result` (Result) — and D13 wants them unified. That is ~94 call sites and the duplication removal is the real work.

Both are worth their own PR rather than being bolted onto this one.

## Merge timing

### DIFF
--- a/std/encoding/base64.yo
+++ b/std/encoding/base64.yo
@@ -64,19 +64,19 @@ export(base64_encode, base64_encode_url);
 // ============================================================================
 // Decoding
 // ============================================================================
-_decode_char :: (fn(c : u8, alpha : str, exn : Exception) -> u8)({
+_decode_char_res :: (fn(c : u8, alpha : str) -> Result(u8, EncodingError))({
   i := usize(0);
   while(i < usize(64), i = (i + usize(1)), {
     cond(
       (alpha.bytes(i) == c) => {
-        return(u8(i));
+        return(.Ok(u8(i)));
       },
       true => ()
     );
   });
-  exn.throw(dyn(EncodingError.InvalidChar(c)))
+  .Err(EncodingError.InvalidChar(c))
 });
-_decode_with :: (fn(s : String, alpha : str, exn : Exception) -> ArrayList(u8))({
+_decode_with_res :: (fn(s : String, alpha : str) -> Result(ArrayList(u8), EncodingError))({
   // Strip trailing padding
   len := s.len();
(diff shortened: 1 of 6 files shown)


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #468, opened 2026-09-07
std: ArrayList.push is infallible; try_push is the allocator-aware form (D9)

shd101wyy:
## What

`push -> Result(unit, ArrayListError)` was discarded by **every single one of its callers**, and the discard was usually spelled as a throwaway binding invented purely to swallow a value nobody wanted:

```rust
_r0 := fns.push(String.from(BF_YO_OP_ADD)); // +
_r1 := fns.push(String.from(BF_YO_OP_SUB)); // -
_r2 := fns.push(String.from(BF_YO_OP_MUL)); // *
```

A `Result` nobody checks is strictly **worse** than a panic: at the call site it reads as if the failure were handled. **D9** in `plans/STD_API_STABILIZATION.md` §2.

Meanwhile the list's own neighbours already panicked on the same failure (`ensure_total_capacity`, `with_capacity`), as do `HashMap.new` and `Deque.push_back`. Rust's `Vec::push` aborts. `push` was the odd one out.

- **`push -> unit`**, panicking on allocation failure
- **`try_push -> Result(unit, ArrayListError)`** — the same body, renamed

## The line this draws

> **An allocation failure panics. A caller mistake keeps its `Result`.**

So **`LinkedList.insert -> Result(unit, LinkedListError)` is deliberately untouched** — its error is an out-of-bounds index, which is the caller's bug to handle, not the allocator's.

## Scale

**226** throwaway bindings removed across `std/` + `src/`, plus **13** in `tests/`. Two sites needed thought rather than rewriting:

- **`BTreeMap.insert` performed the push inside a `cond` *condition*** (`push(...).is_err() => __yo_panic(...)`), so the push was a *side effect of testing it*. Moved into the `true` arm — equivalent now that `push` panics itself. The comment says so, because it is not obvious from the diff.
- **`PriorityQueue.push`** had the same guard and now just calls `push`.

## `std/imm/` is deliberately untouched

`ImmVec.push` is a **persistent** vector that returns a *new* vec, so `v2 := v1.push(x)` is correct there and means something completely different. The sweep only removed underscore-prefixed bindings that are **never read**, which is structurally why it could not touch those — but it is worth stating, since a careless regex over `.push(` would have wrecked them.

## Verification is the BUILD, not `check`

`yo check` does **not** evaluate deferred generic bodies, so it passed **173/173 + 266/266 even before the consuming sites were fixed**. The real gates:

- **the compiler compiles itself** with this change (`yo build` green)
- **full suite: 3593/3593** (`tests`, excluding `tests/internal` and `tests/cli-cases`)
- `fmt --check` clean on `./std ./src ./tests`

## D9's `insert` half is NOT here — and here is why

It has **491** consumers rather than push's 281, and it carries a hazard that deserves its own review:

`map.insert(k, v).unwrap()` today unwraps the **`Result`**, giving an `Option(V)`. Under `insert -> Option(V)` the same text unwraps the **`Option`** and **panics whenever the key was absent**.

There are 8 such sites, and they are inside the **`hash_map!` / `hash_set!` literal macros**, which expand into user code — so every fresh-key literal would panic at runtime *while still type-checking*. That is not a footnote to this PR.

## Also in this PR

`AGENTS.md` gains the squash-merge/delete-branch rule (folded in here because a docs-only PR cannot merge without `--admin`).

## Merge timing


shd101wyy:

```
```

| gate | result |
| --- | --- |

PR #474, opened 2026-09-07
std/fmt: derived ToString/Debug bodies reference no name from the deriving file

shd101wyy:
## The bug

A derive body is spliced through `.to_expr()`, which **re-parses** it, so every free
name in it resolves in the *deriving file's* scope — not in `std/fmt/to_string.yo`'s.
`__derive_structural_body` emitted `String.from("P(")`, so a file deriving
`ToString`/`Debug` without `String` in scope hit `E0401: Variable "String" not found`
at definition time, had it **swallowed**, and compiled to:

```c
__attribute__((error("yo: the body of fn_yo_id_7585 failed to transpile — its
  definition-time evaluation failed and was swallowed ...")))
static inline __yo_t1 fn_yo_id_7585(__yo_t0* self) {
  abort(); /* untranspilable body in a value-returning fn */
}
```

`yo check` reported **"evaluator OK"**, and the program died at rc=134 with no
output. The `__attribute__((error(...)))` guard does not fire at `-O2` even though
the function is reached.

Pre-dates D15 — `derive(ToString)` was affected identically. It went unnoticed
because every std module that derives happens to import `String`, as does
`tests/derive.test.yo`.

## The fix

Emit every literal as `"P(".to_string()`. A `str` literal introduces no identifier,
and the method resolves through `impl(str, ToString)` by *receiver type*, never by
scope. The only remaining names — `self`, the `__v_*` arm bindings, the `.`-prefixed
variant tags — are all bound at the splice site. Rendered output is unchanged.

Verified with **no** `String` import in the deriving file:

| shape | before | after |
|---|---|---|
| `derive(P, Debug)` struct | rc=134, 1 FTT stub | `P(1, 2)` |
| `derive(Z, Debug)` empty struct | rc=134 | `Z()` |
| `derive(E, Debug)` payload-free enum | rc=134 | `E.B` |
| `derive(P, ToString)` | rc=134 | `P(1, 2)` |
| `derive(E1, Debug)` enum w/ payload | `E1.NotFound(/tmp/x)` | unchanged |

### Two things measured on the way

**`quote(...)` is only partially hygienic.** A top-level `::` *defined* in the rule's
module resolves from the splice site; an *imported* name (`String`) does not; nothing
inside a `.to_expr()`'d string does. Worth knowing before the next derive rule.

**A template literal would also be identifier-free, but `.to_expr()` can't parse
one** — filed as `issues/comptime-str-to-expr-cannot-parse-a-template-literal.md`,
with the misleading `derive rule must return(comptime(Expr)); got Comptime` it
surfaces as. The cheatsheet already warned that a raw backtick "splits the parse"
and recommended `String.from("x")` — i.e. the very form that caused this bug — so
that guidance is corrected here too.

## Regression test

`tests/cli-cases/derive-render-without-string-import` — a `build run` fixture
deriving in a file with no `String` import, asserting all four rendered lines.
**Verified to FAIL on the pre-fix std** (rc golden=0 run=1, all four stdout lines
missing) and pass after. The unit tree cannot host it: `tests/derive.test.yo`
imports `String`, and `yo check` is green on the broken program either way.

## Also in this PR (the doc fixes it started as)

- `std/fmt/to_string.yo`'s D15 comment showed `derive(Debug);`, which is not the API:
  `derive([generic(...)], [where(...)], Type, Trait1, ...)`, and fewer than 2 args is
  rejected. The tests were always right (`derive(_D15Point, Debug)`); only the prose
  was wrong.
- "all ten std error enums" → 16 error types hand-write `to_string` today.
- Three skill files showed `derive(Point, Eq, ...)`; bare `Eq`/`Ord` are rejected
  ("Argument count mismatch: expected 1, got 0") and `Ord` additionally needs `Eq`
  in the same call. `syntax-cheatsheet.md:1221` already had this right, so the skill
  files contradicted each other. `docs/{en-US,zh-CN}/DERIVE_TRAITS.md` were correct.

## Still open (not fixed here)

`derive` swallowing a definition-time body failure at all. The FTT stub plus a green
`check` is why this survived a release — the more general defect.

## Verification

- `yo check ./std` 173/173, `yo check ./src` 266/266
- `yo fmt --check` clean on `./std` and `./src`
- compiler self-build green
- `derive` 53, `derive_clone_complex` 16, `fmt` 8, `bootstrap_verification` 12 — all passed
- full suite + complete cli-diff scorecard: running locally, results in a comment


shd101wyy:
## Local verification (complete)

| gate | result |
|---|---|
| `yo check ./std` | **173/173** |
| `yo check ./src` | **266/266** |
| `yo fmt --check ./std` / `./src` | clean |
| compiler self-build (`yo build`) | green |
| full suite (`./tests` minus internal/cli-cases) | **3621 / 3621** |
| cli-diff scorecard | **PASS 71 / GOLDEN-DIFF 0 / SKIP 1** (72 total, +1 new case) |
| `tests/http` / `tests/net` | 43/43 · 58/58 |

### On the golden re-records

Six cases (`init`, `init-build-test`, `init-cwd`, `init-existing`,
`skills-install`, `skills-install-zh`) record the content hash of every file
`yo init` / `yo skills install` copies out of `.github/skills`, so editing a
skill doc moves them. The diff names **only** the three files this PR edits —
`yo-core-patterns/{SKILL.md,core-patterns-cheatsheet.md}` and
`yo-syntax/syntax-cheatsheet.md` — and no golden lost output.

Harness note for whoever runs this next: those six need
`YO_SKILLS="$PWD/.github/skills"` when `YO_SELF_BIN` sits outside the checkout.
Without it they fail as *"bundled skill files not found"* with 12 files missing
from the tree, which reads like a regression but is a missing env var.

### One thing I could not pin down

The first full-suite run reported **3 failed / 3618 passed**, and my gate script
kept only the last 6 lines, so the failing test names were lost. A clean re-run
of the same tree and the same binary gave **3621/3621 with zero `✗` markers**,
and `tests/http` (43/43) and `tests/net` (58/58) — the network-dependent files,
which include the already-filed httpbin flake
(`issues/http-limits-test-depends-on-httpbin-org-inside-a-required-gate.md`) —
both pass in isolation. So the three were transient, but I am recording that I
never identified them rather than asserting they were harmless. CI's Linux legs
are the real check on that.

PR #502, opened 2026-09-08
std/string: FromString on every numeric primitive — and the rename that made the name true

shd101wyy:
**Stacked on #501** (→ #500 → `develop`). Base retargets down the chain as each lands.

## What

`FromStr` was implemented for **six** types — `i64`, `u64`, `i32`, `u32`, `f64`, `bool` — so `s.parse(usize)` did not compile at all:

```
error[E0602]: Type usize does not implement required trait FromStr.
```

`usize` is the width that indexes every collection, which makes it the sharpest of the seven gaps. Rust implements `FromStr` for every integer and float primitive; the missing seven (`usize`, `isize`, `i8`, `i16`, `u8`, `u16`, `f32`) are added here.

## How

One generic impl is not possible today — there is no `Integer`-marker-bounded `FromStr` blanket that can name its own width — so each narrows the shared `_parse_i64_radix_res` / `_parse_u64_radix_res` result and reports the **side-specific** overflow variant rather than a truncated value.

`usize` and `isize` take their bounds from the width-aware `MIN`/`MAX` associated constants (already `cond`ed on `__yo_pointer_size_bits()`), not from a literal — so a 32-bit target rejects `4294967296` where a 64-bit one accepts it, without this code knowing which target it is on.

`f32` shares `f64`'s grammar and then narrows, and an over-large literal becomes an **infinity rather than an error**: `"1e39".parse::<f32>()` is `Ok(inf)` in Rust too. The narrowing is C's double-to-float conversion, which saturates per IEC 60559 (C Annex F) on every target here.

## Tests

Per width: a value, the exact `MAX` — round-tripped through `to_string()` so the assertion cannot disagree with the constant it is testing — `MAX + 1` as `PosOverflow`, and for the signed types the exact `MIN` and `MIN - 1` as `NegOverflow`. Plus an empty string, a rejected sign on `usize`, and `f32`'s empty / garbage / saturating cases.

## Also

- `JsonValue.pointer` (added in #501) drops its `parse(u64)`-and-narrow workaround for a plain `token.parse(usize)`.
- `plans/archive/RELEASE_NOTES_v0.2.29_DRAFT.md` — the curated text edited into the [v0.2.29 release](https://github.com/shd101wyy/Yo/releases/tag/v0.2.29), archived beside v0.2.26's and v0.2.28's. Folded in here because a docs-only PR skips the required contexts and cannot merge without `--admin`.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` on both changed files | clean |
| standalone probe of all seven widths at `--optimize 2` | passes, including `1e39` → infinity |

`issues/fixed/fromstr-missing-on-seven-numeric-primitives.md`


shd101wyy:
**Local gates (macOS arm64), on this branch's tip:**

| gate | result |
| --- | --- |
| `yo build` | rc=0 |
| full suite (`./tests`, minus internal/cli-cases) | **3849 passed, 0 failed** |

That suite covers the whole stack — #500's `Default` tests, #501's `JsonValue` tests and this PR's `FromStr` tests all ran in it.

shd101wyy:
## Added: `FromStr` → `FromString`

The maintainer spotted that the trait's name names a type it cannot accept, and they're right — the parameter is a `String`, and the doc comment said so one line above the signature:

```rust
/// `FromStr` — parse a value of this type out of a `String`
FromStr :: trait(Err : Type, from_str : (fn(s : String) -> …));
```

Rust's name is **accurate** — `fn from_str(s: &str)` really does take a `&str` — and the pattern behind it is *"name the trait after the type it converts from"*. Applying that pattern in Yo, where `str` and `String` are different types, gives `FromString`.

**The alternative isn't available.** Making it take a `str` so the name becomes true is closed off: `as_str()` was deleted in the slice rework, no method in `std/string/string.yo` returns `str`, and `String.parse(T)` — the primary caller — has a `String` receiver. `String` is the only parameter this trait can take, so the name is what had to move.

Two other sites made the same mistake and are swept with it:

| before | after |
| --- | --- |
| `log.level_from_str(name : String)` | `level_from_string` |
| `HttpMethod.from_str(s : String)` | `HttpMethod.from_string` |

`std/imm/string.yo` already had `from_string(s : String)` — the tree was inconsistent with **itself**, not just with Rust.

Rust **citations** in doc comments keep Rust's spelling: `core::str::FromStr`, `i64::from_str`, `from_str_radix` name Rust's API, not Yo's. The rename is written to leave every one alone — 12 survive in `string.yo`, all preceded by `::`.

**Breaking, deliberately without an alias.** A `FromStr :: FromString` alias keeps `where(T <: FromStr)` working but *not* `from_str : …` inside an impl, because a field name is part of the trait. A half-bridge covering one of the two ways a trait is used is worse than a clean rename in the release notes.

Filed rather than folded in: `HttpMethod.from_string` returns `Option`, not `Result` — a D12 violation independent of its name, and an error-type design question rather than a rename (`issues/httpmethod-from-string-returns-option-not-result.md`).

### Gates on the rename

| gate | result |
| --- | --- |
| `yo fmt --check` on all 11 changed files | clean |
| `tests/string/string_parse.test.yo` | 42 passed |
| `tests/string/string.test.yo` | 289 passed |
| `tests/log_sink_env.test.yo` | 8 passed |
| `tests/http/server.test.yo` | 13 passed |
| CLI golden scorecard (after re-recording `lsp-import-members`) | **PASS 88, GOLDEN-DIFF 0, NO-GOLDEN 0** |

The golden diff is exactly the one completion label `FromStr` → `FromString`; nothing lost. Full suite re-running.

shd101wyy:
Full suite on the tip (rename included): **3855 passed, 0 failed**. Fixpoint running.

shd101wyy:
`fixpoint_only.sh` on the tip: **FIXPOINT_HOLDS** (stage2 hollow=0, CLANG_RC=0, STAGE3_RC=0). Local coverage complete.

PR #503, opened 2026-09-08
std/math: INFINITY, NEG_INFINITY and NAN — and the MSVC NaN spelling that fell out

shd101wyy:
The last three names `std/math` was missing — **unblocked by the v0.2.29 seed**, which landed a few hours ago.

They had to wait for a seed because the v0.2.28 compiler emitted a non-finite float constant as the invalid C literal `inf.0` (`issues/fixed/comptime-float-infinity-emits-invalid-c.md`, fixed in #487 and therefore first present in a seed at v0.2.29), and `std/` must build under the seed. Verified against the real v0.2.29 bundle, not the local dev binary.

## The spelling, and the two rejected alternatives

None of the three has a token of its own, so each is spelled as the thing that produces it:

```rust
INFINITY : f64(1.0e400),
NEG_INFINITY : f64(-1.0e400),
NAN : (f64(1.0e400) - f64(1.0e400)),
```

Both alternatives were tried against the actual compiler and rejected:

- `0.0 / 0.0` — refused outright: `Division by zero in comptime float operation: __yo_comptime_f64_div`.
- `x * y - x * y` — `-ffp-contract=on` fuses it into an FMA, and it evaluates to `-inf`, not NaN.

The doc says all of this at the definition, so the `1.0e400` does not read as a typo.

## Tests

They do not stop at `is_infinite()` / `is_nan()`, which would pass even for a constant that never reached C correctly. They assert `INFINITY > MAX`, `NEG_INFINITY < MIN`, `NEG_INFINITY == -INFINITY`, that **NaN is unequal to itself** — the whole reason `is_nan()` exists, and why the doc says never to write `== f64.NAN` — and that arithmetic on the constants behaves: `inf + 1`, `inf * -1`, `1 / inf`, NaN propagation. That last group is what would have caught `inf.0`.

Same coverage at `f32`.

## Stability

`std/math`'s `## Stability` marker drops these from its list of deliberate gaps. The one that remains is the `f32` transcendentals C has no single-precision entry point for.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` | clean |
| `yo check ./std` (v0.2.29 seed) | 174/174 |
| `tests/math.test.yo` | **12 passed** (10 + 2 new) |

### DIFF
--- a/issues/fixed/msvc-nan-spelling-escapes-the-non-finite-float-match.md
+++ b/issues/fixed/msvc-nan-spelling-escapes-the-non-finite-float-match.md
@@ -1,0 +1,88 @@
+# The MS CRT's `-nan(ind)` escapes the non-finite float match, emitting `-nan(ind).0`
+
+**Status: FIXED** (2026-09-09) — `src/codegen/exprs/comptime_value.yo`.
+
+## Symptom (Windows only)
+
+A comptime NaN constant reached the C compiler as:
+
+```c
+bool _file___D__a__temp_17067 = yo_id_10435(-nan(ind).0);
+```
+
+```
+tests/.yo_selftest_batch_139_0.bin.c:5652:78: error: use of undeclared identifier 'ind'; did you mean 'bind'?
+ 5652 |                             bool _file___D__a__temp_17067 = yo_id_10435(-nan(ind).0);
+      |                                                                              ^~~
+      |                                                                              bind
+C:\...\um\winsock2.h:1655:1: note: 'bind' declared here
+```
+
+macOS and Linux were green on the same commit. The suggestion of `bind` from
+`winsock2.h` is a coincidence of the payload's spelling, and a good reminder
+that a C-level diagnostic can point somewhere unrelated to the cause.
+
+## Root cause
+
+`generate_comptime_value`'s `.FloatLit(raw)` arm turns the three non-finite
+doubles into their C11 spellings (`HUGE_VAL`, `(-HUGE_VAL)`, `NAN`) before the
+"append `.0` if there is no radix point" rule can run — that rule is what
+produced `inf.0` in
+`issues/fixed/comptime-float-infinity-emits-invalid-c.md`.
+
+It recognised them by **string equality**:
+
+```rust
+((lowered == "nan") || (lowered == "-nan")) => …
+```
+
+But the raw comes from the HOST C library's `%g`, and the libraries disagree:
+
+| library | NaN renders as |
+| --- | --- |
+| glibc, Apple libc, musl | `nan`, `-nan` |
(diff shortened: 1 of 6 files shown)


shd101wyy:
The `test (windows-latest)` failure was **real and mine**, not a flake — pushed a second commit fixing it.

```
tests/.yo_selftest_batch_139_0.bin.c:5652:78: error: use of undeclared identifier 'ind'; did you mean 'bind'?
 5652 |   bool _file___D__a__temp_17067 = yo_id_10435(-nan(ind).0);
```

`generate_comptime_value` turns the three non-finite doubles into their C11 spellings before the "append `.0` if there is no radix point" rule can run — that rule is what produced `inf.0` in the bug this PR's constants were waiting on. But it recognised them by string **equality** against `"nan"` / `"-nan"`, and the raw is the **host C library's** `%g` output:

| library | NaN renders as |
| --- | --- |
| glibc, Apple libc, musl | `nan`, `-nan` |
| MS CRT | `-nan(ind)`, `nan(snan)` |

So the equality test passed on every platform a developer looks at and failed on the one that spells the payload. `f64.NAN` is `∞ - ∞`, which on x86 is precisely the indefinite NaN with the sign bit set — the `-nan(ind)` case.

Matched by **prefix** now, `+` sign accepted. No finite raw can collide: `%g` renders one starting with a digit, `-`, `+` or `.`, never with `i` or `n` after an optional sign. The NaN's sign and payload are dropped on purpose — C has no portable constant expression for either, `NAN` is the canonical quiet NaN, and nothing but `copysign`/`signbit` can observe the difference.

**There is no way to exercise this from a macOS or Linux host**: the raw comes from the host's `%g` at comptime, not from the target, so a cross-emit does not reproduce it. `tests/math.test.yo`'s two non-finite tests *are* the reproducer, and the Windows leg is the gate.

Filed the std-level half separately — `f64.NAN.to_string()` is likewise platform-spelled where Rust prints `NaN` everywhere, and that also decides what `json_stringify` should do with a NaN, since JSON has no non-finite literal (`issues/float-to-string-is-platform-dependent-for-non-finite-values.md`).

shd101wyy:
**Local gates (macOS arm64), on the rebased tip (both commits):**

| gate | result |
| --- | --- |
| `tests/math.test.yo` | **12 passed** (10 + 2 new) |
| full suite (`./tests`, minus internal/cli-cases) | **3840 passed, 0 failed** |

None of that can see the Windows failure this PR's second commit fixes — the raw string comes from the **host's** `%g` at comptime, not from the target, so no macOS or Linux gate and no cross-emit reproduces the MS CRT's `-nan(ind)`. `test (windows-latest)` is the gate for that half, and it is why the first push of this PR was red.

PR #504, opened 2026-09-08
std/net: Eq/Ord/Hash/Clone on IpAddr and SocketAddr; TcpStream.local_addr

shd101wyy:
## Traits on the two address types

§1's trait-coverage row read *"`Eq`/`Hash` on 0/2 net address types"*. Both now have `Eq`, `Ord`, `Hash` and `Clone` — which is what makes a connection table keyed by peer, or a sorted peer list, possible at all. The test asserts exactly that use, including that re-inserting an equal-but-distinct key **replaces** rather than duplicates.

All four are hand-written rather than derived, because the order Rust gives these types is **not** the structural one:

- `Ipv4Addr` compares as its 32-bit value and `Ipv6Addr` as its 128-bit one, so `1.255.255.255 < 2.0.0.0`. A field-by-field derive would also give that, but only by accident of field order.
- **Every V4 address sorts before every V6 one.** Rust gets that from deriving `Ord` over `V4 < V6` in declaration order; here it is written down and tested (`255.255.255.255 < ::1`).
- `Hash` feeds a one-byte **family tag** first, so a V4 and a V6 address cannot collide by sharing a byte pattern.

`SocketAddr` orders IP first, port second — Rust's structural order for a derived pair — and delegates both halves.

## `TcpStream.local_addr`

Reads the kernel's answer **once**, at connect / accept time, and stores it beside `_peer_addr`. The ephemeral source port and the source interface the kernel picked are not in the `connect` argument, so they have to come from `getsockname`; storing the result keeps the accessor infallible and symmetric with `peer_addr`.

The `getsockname` read-back that `TcpListener.bind` open-coded is now a `_read_local_addr` helper, whose doc says when the fallback can fire and why `bind`'s fallback (the bind argument) is meaningful where a connected socket's is not.

The test is the one that actually proves it: the client's `local_addr` must be a loopback address with a non-zero port that is **not** the destination port, the server's `peer_addr` must equal it, and the accepted socket's `local_addr` must be the listening port.

## `TcpListener.incoming` is deferred, with the reason recorded

Rust's `incoming()` is a **blocking** iterator of `io::Result<TcpStream>`. Yo's `accept` is `Impl(Future(TcpStream, IoExn))`, and there is no `Stream` trait — no async analogue of `Iterator` — for an iterator of futures to implement. An `Iterator` that blocks per element would be worse than nothing: a blocking await inside a task nests the event loop and deadlocks. The row wants an async-iteration abstraction first; it is a design question, not a missing method. Written up in the plan.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` | clean |
| `yo check ./std` (v0.2.29 seed) | 174/174 |
| `yo build` | rc=0 |
| `tests/net/addr.test.yo` | **20 passed** (4 new) |
| `tests/net/tcp.test.yo` | **18 passed** (1 new, over a real loopback connection) |

### DIFF
--- a/issues/yo-fmt-walks-gitignored-generated-files.md
+++ b/issues/yo-fmt-walks-gitignored-generated-files.md
@@ -1,0 +1,45 @@
+# `yo fmt` walks `.gitignore`d files, so `--check ./tests` fails on test-runner scratch
+
+**Status: OPEN.** Found 2026-09-09.
+
+## Symptom
+
+```
+$ yo fmt --check ./tests
+The following Yo files need formatting:
+tests/string/.yo_selftest_batch_192_0.yo
+```
+
+`.yo_selftest_batch_*` is the test runner's own scratch — the file it inlines
+every test body of one file into before compiling the batch. It is
+`.gitignore`d (`.gitignore:39`), it is machine-generated, and it is never meant
+to be formatted or read by a human.
+
+The formatter's directory walker does not consult `.gitignore`, so any tree
+where a test run has been interrupted — or is running right now — reports a
+false formatting failure. It bit a local gate script whose first step is
+`yo fmt --check ./tests`; the "failure" was a leftover from a run killed twenty
+minutes earlier.
+
+CI does not see it because it formats before it tests, so this is a local-only
+papercut — which is exactly the kind that wastes a debugging cycle, because the
+named file does not exist in `git status`.
+
+## Two candidate fixes
+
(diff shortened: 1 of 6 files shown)


shd101wyy:

| gate | result |
| --- | --- |
| full suite (`./tests`, minus internal/cli-cases) | **3820 passed, 0 failed** |

PR #505, opened 2026-09-08
std: a `## Stability` marker on the six modules that lacked one

shd101wyy:
§1's last audit row counted **eight** modules with no `## Stability` section — the thing `yo doc` renders and readers use to tell a settled name from a provisional one. `std/http/server` picked one up when the server was last touched; the other six get one here: `std/async/index`, `std/async/channel`, `std/async/mutex`, `std/sync/barrier`, `std/sync/semaphore`, `std/gc`.

Each says **what is expected to move and what is not**, rather than a bare stable/unstable word:

- The three `async` modules are unstable in their **mechanism**, not their names — every one of them re-checks a blocked operation on a 1 ms timer tick, which a waker-based rewrite would remove. That changes latency and CPU cost, not a signature.
- `sync/barrier` and `sync/semaphore` are **stable**: each follows a named model exactly (the generation barrier that `std::sync::Barrier` and `pthread_barrier_t` share; Java's and `tokio`'s counting semaphore), and the two properties most likely to be misread as provisional — unfair acquisition, an uncapped permit count — are deliberate and already documented.
- `std/gc` also gains **the module doc it never had**. It had one line. It now says what the cycle collector is for (reference counting reclaims everything except a cycle) and that `tracked_count` counts cycle *candidates* rather than live objects, because a type proven `Acyclic` never reaches the collector at all.

## Two audit rows settled while writing them

- **`async/mutex.with_lock` already takes an `io`**, so §4's *"either takes an `io` so its doc claim becomes true, or drops the claim"* is satisfied as written. The claim is true today; no change needed.
- **`async/channel.try_recv` still returns `Option(T)`** where `sync/channel`'s now returns `Result(T, TryRecvError)` (#495). The marker names that divergence as the thing that will move. Aligning it is a follow-up — a signature change, not a doc change.

Both recorded in `plans/STD_API_STABILIZATION.md`.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` on all six | clean |

Doc-comment-only in `std/*.yo`, so no behaviour changes — but they are `.yo` files, so this is not on the docs-only fast path and gets the full matrix.

### DIFF
--- a/plans/STD_API_STABILIZATION.md
+++ b/plans/STD_API_STABILIZATION.md
@@ -647,6 +647,28 @@ rewritten over `Mutex.with_lock` (its blocker is in `issues/fixed/`);
    (issues/mutual-recursion-between-a-fn-and-a-trait-impl-body.md, with a
    minimal reproducer). `Eq` is therefore ONE self-recursive `_json_eq`, which
    is what `Clone` already did.
+   **`## Stability` markers: DONE (2026-09-09).** §1's last row counted eight
+   modules without one. `std/http/server` got its marker when the server was
+   last touched; the remaining six now have one — `std/async/index`,
+   `std/async/channel`, `std/async/mutex`, `std/sync/barrier`,
+   `std/sync/semaphore` and `std/gc` (which also gains the module doc it never
+   had: what the cycle collector is FOR, and that `tracked_count` counts cycle
+   CANDIDATES, since an `Acyclic` type never reaches the collector).
+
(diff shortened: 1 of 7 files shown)


shd101wyy:

| gate | result |
| --- | --- |
| full suite (`./tests`, minus internal/cli-cases) | **3815 passed, 0 failed** |
| `fixpoint_only.sh` | **FIXPOINT_HOLDS** |

Doc comments only, so the interesting result is the boring one: nothing moved.

PR #506, opened 2026-09-08
std/async: try_recv returns Result(T, TryRecvError), matching the blocking channel

shd101wyy:
**Stacked on #505.** Base retargets to `develop` once #505 lands.

## What

`std/sync/channel.try_recv` moved to `Result(T, TryRecvError)` in #495, because `Option(T)` collapsed the two answers a polling caller must act on differently: `.None` meant both *"nothing buffered yet, retry"* and *"closed and drained, no value will ever arrive"*.

The **async** channel was left behind, so the two channels reported the same distinction in two vocabularies. #505's stability marker had to write that divergence down as a known defect; this closes it.

It now returns the **same** `TryRecvError`, imported from `std/sync/channel` rather than redeclared, and re-exported so a caller of only the async channel can name it.

## The test that the old shape could not express

```rust
ch := Channel(i32).new(usize(2));
_s := ch.try_send(i32(7));
ch.close();
// A closed channel still yields what it buffered:
assert(ch.try_recv() is Ok(7));
// Only the call AFTER that is Disconnected:
assert(ch.try_recv() is Err(Disconnected));
```

With `Option(T)` both of those last two states were `.None`.

## Stability marker

Drops the divergence and keeps the one question that remains: whether `recv`'s `Option(T)` — where `.None` means closed — should become a `Result` for the same reason.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` | clean |
| `tests/async/channel.test.yo` | **5 passed** (1 new) |

### DIFF
--- a/plans/STD_API_STABILIZATION.md
+++ b/plans/STD_API_STABILIZATION.md
@@ -813,6 +813,13 @@ rewritten over `Mutex.with_lock` (its blocker is in `issues/fixed/`);
    `Result` — a D12 violation independent of the name, and an error-type design
    question rather than a rename
    (issues/httpmethod-from-string-returns-option-not-result.md).
+   - **`async/channel.try_recv` returned `Option(T)`** where `sync/channel`'s
+     returns `Result(T, TryRecvError)` (#495). ALIGNED in the follow-up: the
+     async channel now returns the SAME `TryRecvError`, imported from
+     `std/sync/channel` rather than redeclared, so the two channels report one
+     distinction with one vocabulary. A closed channel with buffered values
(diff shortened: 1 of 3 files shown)


shd101wyy:
**Local gates (macOS arm64), on the rebased tip:**

| gate | result |
| --- | --- |
| `tests/async/channel.test.yo` | **5 passed** (1 new) |
| full suite (`./tests`, minus internal/cli-cases) | **3853 passed, 0 failed** |
| `fixpoint_only.sh` | **FIXPOINT_HOLDS** |

PR #507, opened 2026-09-08
codegen: a tuple type gets a named C struct tag and a forward declaration

shd101wyy:
## Symptom

`ArrayList(Tuple(String, String))` does not compile:

```
tup.c:389:3: error: unknown type name '__yo_t2'
```

`yo check` is green — this is a C-compile failure, so it is loud — but it makes `ArrayList(Tuple(K, V))` unusable, which is the natural return type for anything yielding pairs.

## Root cause

`generate_type_declarations` emits a forward declaration for every future-trait, struct and enum type:

```c
typedef struct __yo_t0_struct __yo_t0; // Forward declaration
```

Tuples got none — and **could not have one**, because `generate_tuple_declaration` emitted an *anonymous* struct, which has no tag to declare ahead of its definition. So when a struct holding a **pointer** to the tuple was emitted first — an `ArrayList(Tuple(..))`'s `_ptr` is exactly that shape — it named a type C had not seen yet:

```c
struct __yo_t0_struct { //  : ArrayList(Tuple(0 : String, 1 : String))
  __yo_ref_header_t header;
  __yo_t2* _ptr;                   // line 389
  ...
};
...
typedef struct { ... } __yo_t2;    // line 438
```

Reordering alone would not be a complete fix: a pointer cycle (`A` holding a `Tuple(*B)` while `B` holds an `*A`) has no valid order. C's rule is that a pointer to an incomplete type is fine **provided the name is declared**, and only a named struct tag can be declared ahead of its definition.

## Fix

1. `generate_tuple_declaration` emits `struct <c_name>_struct { … };` — the same shape `generate_struct_declaration` and `generate_enum_declaration` already use.
2. The forward pass gains an `is_tuple_type` arm.

Every use site still names the typedef, so nothing else moves: compound literals (`(__yo_t2){ ._0 = …, ._1 = … }`), member access and pointer arithmetic are unchanged. The emitted C for every tuple changes **shape**, which is why the fixpoint is the acceptance test here rather than byte identity against develop.

## Test

`tests/basic.test.yo` — "a tuple element type inside a generic container declares its C name in time". On the v0.2.29 compiler the whole file fails to C-compile with two `unknown type name` errors, taking its entire batch with it; with the fix, 47/47.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check ./src` | clean |
| `yo build` | rc=0 |
| repro `.yo` | 2 C errors before → 0 after, runs |
| `tests/basic.test.yo` | **47 passed** (1 new) |

## Found by

Writing `Url.query_pairs() -> ArrayList(Tuple(String, String))` for the §4 Encoding row.

`issues/fixed/tuple-type-has-no-forward-declaration.md` + `issues/repros/`

### DIFF
--- a/issues/fixed/tuple-type-has-no-forward-declaration.md
+++ b/issues/fixed/tuple-type-has-no-forward-declaration.md
@@ -1,0 +1,93 @@
+# A tuple type has no forward declaration, so `ArrayList(Tuple(..))` emits an unknown type name
+
+**Status: FIXED** (2026-09-09) — `src/codegen/types/generation.yo`.
+
+## Reproducer
+
+`issues/repros/arraylist-of-tuple-unknown-type-name.yo`:
+
+```rust
+{ ArrayList } :: import("std/collections/array_list");
+open(import("std/string"));
+main :: (fn() -> unit)({
+  xs := ArrayList(Tuple(String, String)).new();
+  xs.push((`a`, `b`));
+  p := xs(usize(0));
+  cond((p.0 == `a`) => (), true => ());
+});
+export(main);
+```
+
+```
+tup.c:389:3: error: unknown type name '__yo_t2'
+```
+
+`yo check` is green; this is a C-compile failure, so it is loud — but it makes
+`ArrayList(Tuple(K, V))` unusable, which is the natural return type for
+anything yielding pairs (`Url.query_pairs` is what hit it).
+
+## Root cause
+
+`generate_type_declarations` emits a forward declaration for every
+future-trait, struct and enum type:
+
+```c
+typedef struct __yo_t0_struct __yo_t0; // Forward declaration
+```
+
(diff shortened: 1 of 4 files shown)


shd101wyy:

| gate | result |
| --- | --- |
| `yo fmt --check ./src` | clean |
| `yo build` | rc=0 |
| repro `.yo` | 2 C errors → 0, runs |
| full suite (`./tests`, minus internal/cli-cases) | **3816 passed, 0 failed** |
| `fixpoint_only.sh` | **FIXPOINT_HOLDS** (stage2 hollow=0, CLANG_RC=0, STAGE3_RC=0) |

Byte identity against develop is deliberately NOT the gate here — the change alters the emitted shape of every tuple declaration on purpose. The fixpoint (stage2 == stage3, both from the new compiler) plus the full corpus compiling and passing is what makes the case. The full-corpus hollow sweep runs in CI.

PR #508, opened 2026-09-08
std/url: join (RFC 3986 §5.3), query_pairs and path_segments

shd101wyy:
**Stacked on #507**, which this needs: `query_pairs` returns `ArrayList(Tuple(String, String))`, and that did not compile before it.

## `join`

§4's first Encoding row, including the parenthetical that explains why it mattered: *"http hand-rolls redirect resolution because `join` is missing"*.

RFC 3986 §5.3 in full — `_split_ref` decomposes any URI reference into the Appendix-B five components, `_remove_dot_segments` is §5.2.4, `_merge_paths` is §5.2.3, and recomposition is handed back to the **one** parser rather than a second copy of the authority rules.

**All 20 §5.4.1 normal examples and all 12 §5.4.2 abnormal ones are tests**, plus three cases the RFC's table does not cover but a reader will ask about:

- an empty interior segment survives (`/x//y`);
- a colon **after** the first slash is not a scheme (`p/q:r` → `/a/p/q:r`);
- a colon **before** any slash is (`mailto:a@b` replaces everything).

## Three real redirect bugs it fixes

`http/client._resolve_location` now calls `join`. The paste-the-origin version it replaces got these wrong — each verified against the old code before it was written down:

| `Location` | old result | correct |
| --- | --- | --- |
| `../x` from `/a/b/c` | `/a/b/../x` verbatim | `/a/x` |
| `//host/p` | an absolute **path** on the current host | a protocol-relative reference to `host` |
| `?q=1` | the base's **directory** + query | the whole path + query |

The `../` case has an **end-to-end loopback test**: the server answers 404 for the un-normalized path, so the old behaviour fails it. The other two are pinned directly on `join`, because reaching another host or asserting on a query needs no round trip.

## `query_pairs` / `path_segments`

`query_pairs` follows the `application/x-www-form-urlencoded` rules a `?a=b&c=d` query actually obeys: `+` is a space decoded **before** the percent escapes — the other order turns an encoded plus into a space — only the **first** `=` splits, a bare key has an empty value, an empty pair between two `&`s is skipped, and order and duplicates survive.

`path_segments` is `.None` for a URL that cannot be a base, as Rust's is: a `mailto:` path is opaque, not a `/`-separated hierarchy, and splitting it would invent structure that is not there. A trailing `/` yields a final empty segment, which is how `/a/` stays distinguishable from `/a`.

Both decode **lossily**. `percent_decode` returns a `Result`, which is right for a caller decoding one value on purpose and wrong for an accessor over a URL that already parsed — one bad escape in one query value must not turn the whole accessor into an error. Rust's equivalents are lossy for the same reason.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` | clean |
| `tests/url/url.test.yo` | **34 passed** (8 new, incl. all 32 RFC §5.4 examples) |
| `tests/http/http_limits.test.yo` | **6 passed** (1 new, over a real loopback server) |
| full suite + CLI + fixpoint | queued behind #507's; results in a comment before merge |

### DIFF
--- a/plans/STD_API_STABILIZATION.md
+++ b/plans/STD_API_STABILIZATION.md
@@ -739,6 +739,49 @@ rewritten over `Mutex.with_lock` (its blocker is in `issues/fixed/`);
    (issues/float-to-string-is-platform-dependent-for-non-finite-values.md);
    it also decides what `json_stringify` should do with a NaN, since JSON has
    no non-finite literal.
+   **Encoding — `Url.join` / `query_pairs` / `path_segments`: DONE
+   (2026-09-09).** §4's first Encoding row, including the parenthetical that
+   explains why it mattered: "http hand-rolls redirect resolution because
+   `join` is missing".
+
+   `join` is RFC 3986 §5.3 in full — a `_split_ref` that decomposes any URI
+   reference into the Appendix-B five components, `_remove_dot_segments`
+   (§5.2.4), `_merge_paths` (§5.2.3), and recomposition handed back to the ONE
+   parser rather than a second copy of the authority rules. All 20 §5.4.1
+   normal examples and all 12 §5.4.2 abnormal ones are tests, plus the three
+   cases the RFC's table does not cover but a reader will ask about: an empty
+   interior segment survives, a colon after the first slash is not a scheme,
+   and a colon before any slash is.
+
+   `http/client._resolve_location` now calls it, which fixed three real
+   redirect bugs the paste-the-origin version had (each verified against the
+   old code before the claim was written down): `../x` was pasted VERBATIM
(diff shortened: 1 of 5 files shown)


shd101wyy:
**Local gates (macOS arm64), on the rebased tip:**

| gate | result |
| --- | --- |
| `tests/url/url.test.yo` | **34 passed** (8 new — all 32 RFC 3986 §5.4 examples) |
| `tests/http/http_limits.test.yo` | **6 passed** (1 new, over a real loopback server) |
| full suite (`./tests`, minus internal/cli-cases) | **3852 passed, 0 failed** |

PR #511, opened 2026-09-09
std/fmt: pad by RUNES, retire Writer into StringBuilder, export both from std/fmt

shd101wyy:
Two §4 Text rows that turned out to be the same row.

## `write_padded` measured width in BYTES

```rust
// std/fmt/writer.yo — BYTES
len := s.len();                    // str.len() is __yo_str_len

// std/fmt/spec.yo — RUNES
len := text.chars().count();       // "`width` is a CHARACTER count", citing Rust
```

So `{:8}` and `write_padded(s, 8, …)` disagreed on any non-ASCII text. `"héllo"` is **6 bytes and 5 runes**, so it got two pad characters where a column needs three — a table came out ragged exactly on the rows with accents.

It now counts runes by scanning for UTF-8 lead bytes. A scan rather than a decode: `str` has no rune iterator (it exposes `len`, `ptr`, `bytes(i)`), and counting lead bytes needs neither a validation pass nor an allocation.

**Why it survived:** `write_padded` had no caller and no test anywhere in `std`, `src` or `tests` — exported, documented and dead. ASCII also hides the defect completely, since there the two counts agree.

## "One string builder" — resolved by measuring, not choosing

| | references | importers in `std` | methods with a caller |
| --- | --- | --- | --- |
| `StringBuilder` | 87 across 14 files, mostly the compiler | — | — |
| `Writer` | — | **3** | **6 of 18** |

So `Writer` retires **into** `StringBuilder`, not the other way round. Only the methods with a caller came across (`write_hex`, `write_f64`) plus the audit's `write_padded`.

`write_bool`, `write_octal`, `write_binary`, `write_bytes`, `write_i64`, `write_u64` had **none** and are not carried — `sb.push_string(n.to_string())` covers them, and interpolation covers the rest. Carrying nine dead methods to a new home would be the odd choice in a PR whose subject is retiring the type they live on.

`StringBuilder.to_string` inherits `Writer`'s **O(1) hand-over**. It used to allocate a second buffer and copy every byte across through a bounds-checked `get(i)` before resetting, so building an N-byte string cost O(N) again to read back. The detach-semantics test moved with it, so the property that matters — a later write never mutates a `String` already handed to a caller — stays pinned.

## `Alignment` exported

`Alignment` moves to `std/string/string_builder` (where `write_padded` now lives), and both it and `StringBuilder` are re-exported from `std/fmt`. That's the audit's row: it *was* exported — from its own module, and from nowhere a `std/fmt` user would look. The dependency runs one way only; `std/fmt` already imports `std/string`, so the reverse would be a cycle.

## Breaking

`import "std/fmt/writer"` is gone. The three importers in `std` are migrated here. An external caller replaces `Writer` with `StringBuilder` and drops chaining — `StringBuilder`'s writers return `unit`.

## The end state, decided but not done here

`String` is **already** an amortized builder: `push_str` / `push_string` / `push_rune` / `push_byte` append in place over an `ArrayList(u8)`, with `with_capacity` and `reserve`. That's Rust's answer — you build into a `String`. `StringBuilder` should collapse into it, and that's 87 call sites in the compiler, so it gets its own PR and its own gates. Recorded in the plan.

## Local gates (macOS arm64)

| gate | result |
| --- | --- |
| `yo fmt --check` on all changed files | clean |
| `yo check ./std` | 173/173 (one file fewer — `writer.yo` is gone) |
| `tests/fmt.test.yo` | **11 passed** (3 new: rune padding, `write_hex`, migrated detach test) |
| `tests/string/string_builder.test.yo` | 22 passed |
| `tests/net/addr.test.yo` | 20 passed |
| `tests/url/url.test.yo` | 34 passed |

`issues/fixed/write-padded-counted-bytes-where-formatspec-counts-runes.md`
`issues/fixed/writer-to-string-shares-its-buffer.md`


shd101wyy:
### Local gate — green

| gate | result |
| --- | --- |
| `yo fmt --check ./std` | clean |
| Full suite (`./tests`, minus `internal` + `cli-cases`) | **3873 passed, 0 failed** — includes the `tests/dyn.test.yo` canary |
| CLI golden scorecard (all 89 cases) | **PASS 88, GOLDEN-DIFF 0, NO-GOLDEN 0**, 1 network SKIP |
| Bootstrap fixpoint (stage2 == stage3, `P=local`) | **FIXPOINT_HOLDS** — `STAGE2_RC=0`, `STAGE3_RC=0`, `CLANG_RC=0`, `stage2 hollow=0` |

The one golden that moved, `lsp-import-members`, is re-recorded in
afb0a309c and is **purely additive**: `std/string` gains `Alignment`
alongside the `StringBuilder` it already exported, and `std/fmt` re-exports
`StringBuilder` + `Alignment` in place of the retired `Writer`. Nothing was
dropped from either completion list. The full scorecard above was re-run
*after* the record, per the "never trust a single re-recorded case" rule.

shd101wyy:
### Merged `develop` in (#509 landed meanwhile) and re-verified the combination

The branch went out of date when #509 (recursive value enums + aggregate
call-arg casts) landed. #509 is a **codegen** change and this PR's full suite
ran without it, so rather than admin-merging past the staleness I merged
`develop` in, rebuilt, and re-ran the surface where the two could interact:

| file | result |
| --- | --- |
| `tests/fmt.test.yo` | 11/11 |
| `tests/format_specs.test.yo` | 13/13 |
| `tests/string/string.test.yo` | 289/289 |
| `tests/net/addr.test.yo` | 20/20 |
| `tests/url/url.test.yo` | 34/34 |
| `tests/async_value_struct_param.test.yo` (#509's own new test) | 1/1 |

All zero failures. The pre-merge full CI on this branch was all-pass, and the
pre-merge local gate was suite 3873/0, scorecard 88/0, `FIXPOINT_HOLDS`.

PR #513, opened 2026-09-09
std/error: derive(Error) — one message per variant emits ToString AND Error

shd101wyy:
D15's second half. An error enum needed **two** hand-written blocks — a
`ToString` impl whose `to_string` matches every variant, and a bare
`impl(T, Error())` — because `derive(ToString)` emits a STRUCTURAL render
(Rust's `Debug`), which is not a user-facing message. All ten std error enums
paid that twice.

```rust
derive(JsonError, Error(
  .UnexpectedChar => `unexpected character at position ${pos}`,
  .UnexpectedEnd  => `unexpected end of input`,
  .Other          => `JSON error: ${msg}`
));
```

The message is **ordinary Yo**, spliced into the arm that binds the payload, so
`${pos}` is the variant's own field rather than thiserror's positional `{0}` —
nothing new to learn, and a typo in a field name is an ordinary
unresolved-name error.

`JsonError` is migrated here as the proving case: its rendered messages are
byte-identical to the hand-written ones (68/68 `tests/encoding/json.test.yo`)
and the now-redundant `impl(JsonError, Error());` is gone. The remaining
eleven std error enums follow in a stacked PR.

### Shape

The rule builds `impl(T, ToString(…), Error())` directly rather than through
`ctx.make_impl`, which wraps exactly ONE trait body — and two is the whole
point, since `Error` requires `Self <: ToString`. Generic error enums are out
of scope for the same reason (`make_impl` is where the forall/where plumbing
lives, and no std error type is generic).

Messages go in **declaration order**, and each is verified against the variant
it lands on, so a renamed, added, removed or reordered variant is a compile
error. Three `comptime_expect_error` blocks pin that.

### A compiler bug found on the way

Keyed lookup — any order, matched by name — is what this wanted to be, and
`Expr` equality (`ComptimeEq` / `__yo_expr_eq`) makes it expressible. Composing
it inside a rule hits a diagnostics bug, filed here as
`issues/derive-swallows-the-rule-error.md`: `derive` reads the rule's outcome
out of the **expr-info table** rather than catching its exception, so

- a rule's own `comptime_assert` message is replaced by
  `derive: derive rule function failed`, and
- when the error is raised in a `::` binding it is discarded **entirely** —
  `check` and `compile` both exit 0 on a `derive` that produced nothing.

Two minimal repros (`issues/repros/derive-rule-error-*.yo`) and a fix sketch
are in the issue. When it is fixed the ordering constraint can be lifted
without touching a single call site, since declaration order is a valid keyed
list.

### Gate

- `yo check ./std` — 174/174 with the **v0.2.29 seed**, so no seed gate.
- `tests/encoding/json.test.yo` — 68/68.
- `tests/derive.test.yo` — 57/57 (3 new tests + 3 `comptime_expect_error`).
- Full suite + fixpoint results posted as a comment.


shd101wyy:
### Local gate — green

| gate | result |
| --- | --- |
| `yo check ./std` with the **v0.2.29 seed** | 174/174 — no seed gate, the seed already ships everything the rule uses |
| `yo fmt --check` on the three touched `.yo` files | clean |
| Full suite (`./tests`, minus `internal` + `cli-cases`) | **3873 passed, 0 failed** — includes `tests/dyn.test.yo` and both new `derive(Error)` tests |
| `tests/encoding/json.test.yo` | 68/68, messages byte-identical to the hand-written ones |
| Bootstrap fixpoint (stage2 == stage3, `P=local`) | **FIXPOINT_HOLDS** — `STAGE2_RC=0`, `STAGE3_RC=0`, `CLANG_RC=0` |
| CLI golden scorecard | running; result posted below |

No export moved: `__derive_error_head` / `__derive_error` are module-private and
`derive_rule(Error, …)` is a registration, so `std/error`'s export list is
unchanged and no LSP golden shifts.

shd101wyy:
CLI golden scorecard (all 89 cases, in-repo binary): **PASS 88, GOLDEN-DIFF 0, NO-GOLDEN 0**, 1 network SKIP. Local gate complete.

shd101wyy:
### Merged `develop` in (#511 landed) and re-verified the combination

CI on this branch was **all-pass**, then #511 landed and the branch went out of
date. #511 deletes `std/fmt/writer.yo` and reshapes `std/fmt`'s exports, so
rather than admin-merging past the staleness I merged `develop` in
(3fc096ec — the `plans/` edits landed in different paragraphs, no conflict),
rebuilt with `YO_STD=<worktree>/std`, and re-ran the interaction surface:

| file | result |
| --- | --- |
| `tests/derive.test.yo` | 57/57 |
| `tests/encoding/json.test.yo` | 68/68 |
| `tests/error.test.yo` | 14/14 |
| `tests/error_ergonomics.test.yo` | 8/8 |
| `tests/fmt.test.yo` | 11/11 |

All zero failures. The pre-merge local gate was seed `check ./std` 174/174,
suite 3873/0, scorecard 88/0, `FIXPOINT_HOLDS`.

PR #515, opened 2026-09-09
std: every range-taking API is Range-typed, and rand gets a process-global generator

shd101wyy:
Two halves of §4's `rand` row, plus the consistency the first half exposed.

## Ranges are `Range`-typed

`rng.range(i64(1), i64(7))` does not say whether 7 is included.
`rng.range(i64(1) .. i64(7))` does, because `..` **is** the half-open
operator. So the three range-taking APIs in std now take a range value:

```rust
rng.range(i64(1) .. i64(7))            // was range(low, high)
random_range(i64(10) .. i64(20), exn)  // was random_range(min, max, exn)
m.range(k1 .. k2)                      // was range(start, end)
```

`Rng` also gains `range_inclusive(a ..= b)` — a separate method rather than an
overload of `range`, because Yo has no function overloading
(`plans/reference/FUNCTION_OVERLOADING_POLICY.md`), and the alternative (one
`range` over a `SampleRange`-style trait) would put a trait on the public
surface to save one word at the call site. Its full-width case is
special-cased: `end + 1` wraps at `i64.MAX`, so `i64.MIN ..= i64.MAX` draws 64
bits directly. A test asserts that case lands on both signs.

## An empty range: panic for a sampler, empty iterator for a view

`crypto/random.random_range` returned `min` for an empty range — **a value that
was never in the range** — while `Rng.range` panicked for the same input. Two
functions named `*range` disagreeing about the same edge is exactly what this
audit exists to remove, and panicking is the right side: a sampler must return
a value FROM the range.

`BTreeMap.range` keeps yielding an empty iterator for an inverted window, and
now says why in its doc — "no elements" is a real answer where "no number" is
not.

The behaviour change is pinned by a new cli-case
(`crypto-random-empty-range-panics`), mirroring the existing
`rand-empty-range-panics`, because a panic cannot be observed from inside a
test process. The `random_range zero span returns min` test is replaced by a
comment pointing at it.

## A process-global generator — and why it is not called `thread_rng`

`rand_u32`, `rand_u64`, `rand_f64`, `rand_bool`, `rand_below`, `rand_range` and
`rand_range_inclusive` draw from one generator, **lazily** seeded from OS
entropy on first use and serialized behind a `Mutex`. Lazily, because an
entropy read at module-init time would be paid by every program that merely
LINKS `std/rand`; behind a mutex, because the alternative is a data race.

It is deliberately **not** `thread_rng`. Yo has no thread-local storage, so
per-thread generators are not expressible, and the name would promise state the
implementation cannot deliver. The docs say what it is instead — one shared
lock — and tell a hot loop to take its own `Rng.from_entropy()`.

That global is also why `Rng` is now `ref(struct(_st : _RngState))` over a
plain value struct holding the two PCG words: `Rng` itself cannot live in a
`Mutex` (a `ref` type whose refcount is not atomic does not implement `Send`,
and `Mutex(Rng)` is rejected with exactly that message), while `_RngState` can
— and the PCG step stays in exactly ONE place, with `Rng` delegating to it.
`Rng` also gains `next_bool` (off the TOP bit; `x % 2` on a PCG low bit is the
classic weak-bit mistake) and `from_entropy`.

## Gate

| gate | result |
| --- | --- |
| `yo check ./std` with the **v0.2.29 seed** | 174/174 — no seed gate |
| `yo fmt --check` on every touched `.yo` file | clean |
| `tests/rand.test.yo` | 11/11 (6 existing + 5 new) |
| `tests/crypto/random.test.yo` | 10/10 |
| `tests/collections/btree_map.test.yo` | 33/33 |
| Full suite, CLI scorecard, fixpoint | posted below |

The reference-sequence test for (seed 42, stream 54) is **untouched** and still
passes, so restructuring `Rng` did not perturb the generator. New tests cover
both endpoints of `range_inclusive`, the full-i64-width case, `next_bool`'s
balance over 1000 fixed-seed flips, `from_entropy` differing run to run while
`new(seed)` replays, and every global draw staying in its range.


shd101wyy:
### Local gate — green

| gate | result |
| --- | --- |
| `yo check ./std` with the **v0.2.29 seed** | 174/174 — no seed gate |
| `yo fmt --check` on every touched `.yo` file | clean |
| `tests/rand.test.yo` | **11/11** (6 existing + 5 new) |
| `tests/crypto/random.test.yo` | 10/10 |
| `tests/collections/btree_map.test.yo` | 33/33 |
| CLI golden scorecard, re-run **after** recording | **PASS 89, GOLDEN-DIFF 0, NO-GOLDEN 0**, 1 network SKIP (90 cases — the corpus grew by the new panic case) |
| Full suite + fixpoint | running; posted when they land |

Two cli-case goldens were recorded in 33fef268b: `rand-empty-range-panics`'
fixture moved to the `..` spelling so its `src/main.yo` hash moved, and
`crypto-random-empty-range-panics` is new. The hand-written `expected_stdout`
for the new case matched the recording exactly — it was never reported as a
stdout diff, only as a missing tree manifest, so the panic message is what the
`opts` file claims it is.

shd101wyy:
Full suite: **3875 passed, 0 failed** (3873 before + the 5 new rand tests, minus the 3 crypto `random_range` tests folded into the cli-case). All five new tests ran and passed — `range_inclusive covers both endpoints`, `range_inclusive spanning the full i64 width does not overflow`, `next_bool is a fair coin off the top bit`, `from_entropy is unpredictable where new(seed) is reproducible`, `the process-global generator draws in range and is lazily seeded`. Fixpoint running.

shd101wyy:
Bootstrap fixpoint (stage2 == stage3, `P=local`): **FIXPOINT_HOLDS** — `STAGE2_RC=0`, `STAGE3_RC=0`, `CLANG_RC=0`, `stage2 hollow=0`. Local gate complete: seed `check ./std` 174/174, suite 3875/0, CLI scorecard 89 PASS / 0 diff, fixpoint holds.

shd101wyy:
### The `macos-26-intel` failure is a known pre-existing runtime bug, not this PR

That leg failed exactly one test:

```
✗ Test basic spawn of two futures
  Test failed with exit code 6
  counter should be 12 after yield in task2 (at std/assert.yo:25:17)
```

This is **byte-identical** to `issues/yield-resumption-order-...`, filed
2026-09-06 from PR #449 — two tasks that `yield` in submission order resumed in
the reverse order. I did not assume flake; the evidence is:

1. **A prior independent occurrence of the identical assertion**, from a PR
   that touched `std/path`, `std/glob`, `std/encoding/base64`, `std/time`,
   `std/libc/time`.
2. **Neither PR goes near the async runtime.** This one touches `std/rand`,
   `std/crypto/random`, `std/collections/btree_map` and their tests — nothing
   in `std/async`, the runtime, or anything scheduling-related.
3. **The legs swapped.** The first report notes the same head *passed*
   `macos-26-intel` while `macos-latest` failed. Here `macos-26-intel` is the
   one that failed. Both legs can produce either order, which is what "no FIFO
   guarantee" looks like and rules out a runner-image quirk.

e9aa958 records this second sighting in the issue, renames it (the old name
asserted the leg-specificity the second sighting disproves), and states the fix:
`__yo_async_yield` returns a future already in state `-1` (Completed), so the
order is decided by where an awaiting state machine's continuation is queued
and how those deferrals drain — it needs a real FIFO ready-queue. That is a
focused async-runtime change and is **not** bundled in here.

Deliberately NOT done: weakening the assertion to get green. It encodes the
documented cooperative-scheduling contract (`docs/en-US/ASYNC_AWAIT.md`), and a
test that accepted either order would make the contract untestable and hide
future ordering regressions.

### Also in this push

d4d609b fixes a real UB bug I found by inspection while investigating: five
sites computed range spans as a SIGNED subtraction, which overflows i64 for a
wide range (`i64.MIN .. 0` is `0 - i64.MIN`). Signed overflow is undefined
behaviour in C. Now computed unsigned via bit-preserving casts, as Rust's
`gen_range` does. Its commit message is explicit that this is **not** claimed
to be the CI failure and that the new test cannot fail on arm64 — on
two's-complement hardware the wrapped signed result happens to equal the
correct span, so it is a UB removal justified on its own terms.

PR #516, opened 2026-09-09
std/prelude: bit-level integer batteries on all ten integer types

shd101wyy:
§4's Core row asked for `count_ones/leading_zeros/…` and the tree had **none**
of them — only a private SWAR `popcount` hidden inside `std/imm/map.yo`, used
once. This adds the family Rust has:

- `count_ones`, `count_zeros`, `leading_zeros`, `trailing_zeros`,
  `leading_ones`, `trailing_ones` — all ten integer types;
- `rotate_left`, `rotate_right`, `reverse_bits`, `swap_bytes` — all ten;
- `is_power_of_two`, `next_power_of_two`, `checked_next_power_of_two` — the
  unsigned five, as in Rust.

## Why not one generic blanket

Every one of these is WIDTH-dependent, and Yo has no `T.BITS` associated
constant, so `impl(generic(T), where(T <: Integer), T, …)` — the shape the
`checked_*` family uses — cannot express them. They are written **once** over
`u64` (SWAR popcount, bit-smear leading-zeros, `x & -x` trailing-zeros, the
standard reverse/bswap ladders) and delegated per-type with the width as a
literal. That is also how Rust implements them: its macro pastes the width in.

## The signed hazard, and the test that pins it

The signed types bit-cast to their same-width unsigned partner. Yo's integer
conversions are two's-complement bit-preserving, exactly like Rust's `as`, so
`u64(i8(-1))` **sign-extends to sixty-four ones** — the cast must go THROUGH
`u8`, or `i8(-1).count_ones()` silently answers 64 instead of 8. Every signed
delegation is written that way, and a test asserts the count at every width.

## `usize`/`isize` derive their width, not assume it

It is 32 on wasm32 and 64 everywhere else, so a hardcoded 64 would make
`rotate_left` and `reverse_bits` quietly wrong on the wasm legs and nothing
else in the suite would notice. `_USIZE_BITS` reads it off `usize.MAX`, testing
the 32-bit case **first** — because writing `usize(18446744073709551615)` to
test for the 64-bit case would be an out-of-range literal, a hard comptime
error, on exactly the targets it is meant to detect.

## `next_power_of_two` panics instead of wrapping

Rust's release build wraps the overflow to `0`. A silent `0` here is a
zero-sized buffer somewhere downstream, so this panics.
`checked_next_power_of_two` is the form that reports it, matching the
`checked_*` family already present, and is what the tests use since a panic
cannot be observed from inside a test process.

## Cleanup

The private `popcount` in `std/imm/map.yo` is deleted in favour of
`u32.count_ones()`; its single call site now reads
`(bitmap & (bit - 1)).count_ones()`.

## Tests

Nine new tests in `tests/int_checked_arithmetic.test.yo`, whose header widens
to say it now covers both families: complementary one/zero counts at every
width, the negative-signed width case, zero saturating the leading/trailing
counts, rotate-is-not-shift plus rotl/rotr round trips at four widths,
`reverse_bits` and `swap_bytes` as involutions, the power-of-two boundary, and
a `usize`/`isize` test written against `usize.MAX` rather than a literal width
so it stays honest on wasm32.

## Gate

| gate | result |
| --- | --- |
| `yo check ./std` with the **v0.2.29 seed** | 174/174 — no seed gate |
| `yo fmt --check` on every touched `.yo` file | clean |
| `tests/int_checked_arithmetic.test.yo` | **20/20** (11 existing + 9 new) |
| `tests/imm_map`, `imm_set`, `imm_iterators`, `imm_threading` | 22/20/13/30, all green — the `popcount` swap |
| Full suite, CLI scorecard, fixpoint | posted below |

The plan's "Core" paragraph is also corrected: it still claimed the `checked_*`
family and the `f64`/`f32` methods did not exist, both of which landed some
time ago.


shd101wyy:
Full suite: **3881 passed, 0 failed** (3873 before + the 9 new bit-op tests, minus one that merged into an existing block). All nine ran and passed. CLI scorecard and fixpoint running.

shd101wyy:
Correction to the previous comment: the "minus one that merged into an existing block" clause was my own arithmetic on differing per-branch baselines, and it was wrong — disregard it. The verified facts are: `tests/int_checked_arithmetic.test.yo` runs **20 tests in the full-suite log** (11 pre-existing + the 9 added here, all nine named and passing), and the whole suite is **3881 passed, 0 failed**.

shd101wyy:
### A real defect came out of the first scorecard — fixed, not re-recorded

The first CLI scorecard on this branch showed four golden diffs. Three explain
themselves as consequences of the prelude growing by 530 lines / 19 top-level
expressions, and are re-recorded in 63a4b06:

- `check-watch-once`, `check-forward-ref-async-body` — `check: parsed 1184
  top-level exprs` → **1203**, a delta of exactly the 19 added (8 `u64` bit
  helpers + `_USIZE_BITS` + 10 per-type impls).
- `lsp-member-definition` — a prelude definition at line 6864 → **7394**, a
  shift of exactly the 530 lines inserted above it.

**The fourth was not a golden move.** `compile-timeout` printed its diagnostic
**twice**:

```
Yo compilation exceeded the configured time limit (possible evaluator hang). …
Yo compilation exceeded the configured time limit (possible evaluator hang). …
```

Reproduced 3/3 on an idle machine, and the control — the same binary built from
a tree with an unmodified prelude — printed it **once**. So not load, and not
the bit batteries: `mm_preload_prelude` printed the prelude's diagnostic and
then FELL THROUGH, so the compile walked into `mm_eval_entry_exprs`, failed for
the same reason, and printed the rendered error again. Its own comment said "a
prelude that fails to evaluate is the ROOT cause of everything that follows".
The latched evaluator deadline made the second report certain; with the smaller
prelude the 1 ms budget simply expired *after* the prelude finished, so it never
showed.

Fixed in b56a436: `mm_preload_prelude` returns whether the prelude is usable and
the caller acts on it — `compile`/`check`/`verify` exit at the root cause,
while `src/lsp/diagnostics.yo` deliberately keeps degrading, because a language
server must surface the stashed error as a diagnostic rather than die. That is
precisely why the decision cannot live inside the helper.

`compile-timeout` now matches its **original** golden with no re-record, which
is the point: the golden was right and the behaviour was wrong.
`issues/failed-prelude-load-prints-twice-and-continues.md` has the detail.

### Gate after the fix

| gate | result |
| --- | --- |
| `yo check ./src` with the v0.2.29 seed | 271/271 |
| CLI golden scorecard, re-run after the record | **PASS 88, GOLDEN-DIFF 0, NO-GOLDEN 0** |
| Full suite (re-run, since this now touches compiler code) | running |
| Fixpoint | queued |

shd101wyy:
Correction: the golden re-record commit is **38de013**, not 63a4b06 — I mistyped the hash in the comment above. (`b56a436`, the prelude-stop fix, is correct.)

shd101wyy:
Full suite re-run after the prelude-stop fix: **3881 passed, 0 failed**. (The earlier run of this suite was OOM-killed at 2639/3881 — a local machine limit, not a test failure; re-run from scratch.) Fixpoint running.

shd101wyy:
Bootstrap fixpoint (stage2 == stage3, `P=local`): **FIXPOINT_HOLDS** — `STAGE2_RC=0`, `STAGE3_RC=0`, `CLANG_RC=0`, `stage2 hollow=0`.

Local gate complete: seed `check ./src` 271/271, suite **3881/0**, CLI scorecard **88 PASS / 0 diff** (with `compile-timeout` matching its *original* golden, no re-record), fixpoint holds.

PR #517, opened 2026-09-09
std/regex: Rust-shaped names — is_match, find, find_all, one-argument new

shd101wyy:
The regex surface was **JavaScript-shaped** in a standard library that is
otherwise Rust-shaped: `test`, `exec`, `match_all`, `search`, and a
`new(pattern, flags)` whose flag string had to be spelled `` `` `` on every
flagless call — with a sibling `compile(pattern)` existing only to avoid that.

```
re.test(s)        ->  re.is_match(s)
re.exec(s)        ->  re.find(s)
re.match_all(s)   ->  re.find_all(s)
Regex.compile(p)  ->  Regex.new(p)
Regex.new(p, f)   ->  Regex.new_with_flags(p, f)
```

`find_iter` already had its Rust name and is unchanged. One name per shape,
with the common shape getting the short name.

### No separate `captures`

Rust splits `find` (the span) from `captures` (the span plus its groups)
because the first is cheaper. Yo's `RegexMatch` **always** carries its groups —
`group(i)`, `name(n)` — so a second method would return exactly the same value
under a different name. `find`'s doc says this outright, so a reader coming
from Rust does not go looking for it.

### `search` is removed, not renamed

It returned the first match's start as a byte offset, which is
`find(input)`'s `RegexMatch.index()` — a strictly-less-informative duplicate
under a `String.prototype.search` name. The three tests that used it now assert
through `find(...).index()`, which is what they were really checking.

### Blast radius

**One** non-test caller in the entire tree: `src/main.yo`'s
`--test-name-pattern`, now `Regex.new(name_pattern.clone())` and
`nre.is_match(tname.clone())`. Everything else was
`tests/regex/regex.test.yo` (348 call sites) plus doc comments in `index.yo`,
`error.yo` and `flags.yo`.

The test rewrite used a real argument splitter rather than a regex: a naive
``Regex\.new\((.*?),(.*?)\)`` misreads ``Regex.new(`a{2,4}`)`` as a
two-argument call, because the comma is inside the quantifier. The splitter
tracks backtick / double-quote state and bracket depth, so a one-argument call
stays `new` and only a genuine second argument becomes `new_with_flags` — 40
of the 348 sites.

### Gate

| gate | result |
| --- | --- |
| `yo check ./std` with the **v0.2.29 seed** | 174/174 |
| `yo check ./src` with the **v0.2.29 seed** | 271/271 |
| `yo fmt --check` on all five touched files | clean |
| `tests/regex/regex.test.yo`, full suite, CLI scorecard, fixpoint | posted below |


shd101wyy:
### CI complete — 27 checks, all pass

The full matrix went green on head `97f6c360`: the suite on all six targets,
bootstrap fixpoint, the full-corpus hollow sweep, all four `tests/internal`
differential shards, the CLI cases, both wasm legs, ThreadSanitizer and the
pinned-Z3 verification. That is broader coverage than the local gate would
have been, which matters here because the local build was lost to an OOM on my
side and CI did the whole job instead.

### Brought up to date with `develop` and re-verified

#511, #513 and #512 landed while this sat, so `develop` was merged in
(47167efe — clean, no `plans/` conflict). Re-verified on the merged tree:

| gate | result |
| --- | --- |
| `yo check ./src` with the v0.2.29 seed | **271/271** — this is what covers `src/main.yo`'s updated call site |
| `tests/regex/regex.test.yo` | **186/186** |

A note on method: the binary that *runs* a PR's tests does not have to be built
from that PR's tree — `YO_STD` points it at the merged worktree, and the seed's
`check ./src` is what validates a `src/` call-site change. So this needed no
40-minute rebuild.

### One trap worth recording

The first build attempt failed with

```
error[E0603]: Argument count mismatch: expected 2, got 1
     --> src/main.yo:3040:12
```

while `YO_STD=<worktree>/std yo check ./src` was 271/271 on the identical tree.
Cause: `yo build` without `YO_STD` resolves std by walking up from the **seed
binary**, so it compiled `src/` against `~/.cache/yo/versions/0.2.29/std` —
which still has the two-argument `Regex.new`. A green `yo build` in a worktree
is therefore not evidence that `src/` was compiled against that tree's std; it
only bites when `src/` consumes a std API changed in the same tree, which is
exactly what this PR does.

PR #514, opened 2026-09-09
std: every error enum derives its messages instead of hand-writing two impls

shd101wyy:
**Stacked on #513** (`std-derive-error`) — that PR adds the `derive(Error)`
rule; this one uses it everywhere. Will be retargeted to `develop` before
#513's branch is deleted.

The twelve remaining std error enums move from a hand-written `ToString` impl
plus a bare `impl(T, Error());` to a single `derive(T, Error(...))`:
`TimeoutError`, `CryptoError`, `TlsError`, `CsvError`, `EncodingError`,
`RegexError`, `IoError`, `NetError`, `UrlError`, `DateTimeError`,
`HttpParseError`, `HttpError`. `JsonError` came with the rule in #513.

239 lines deleted, 180 added.

### Every rendered message is unchanged, and that is checked, not asserted

The migration was mechanical and verified as such: each existing
`match(self, …)` was split with a balanced-paren splitter, checked arm-for-arm
against the enum's **declaration order**, and the message expression carried
across **verbatim** — including the ones that are not bare templates:

- `String.from("…")` in `IoError` / `NetError`,
- `err.to_string()` for `NetError.Io`,
- the bare `msg` passthrough for `NetError.Other`,
- `` `…`.to_string() `` where the old explicit `-> String` return type needed it.

Every payload binding in the old positional patterns already matched the
declared field name — which is what `derive(Error)` binds by — so no message
changed which field it interpolates.

Two orphaned doc comments (the `DateTimeError` and `TlsError` notes explaining
why the `Error` marker is there) moved onto the derive rather than being
deleted: the marker is still what makes `dyn()` accept them, it is just no
longer a separate line.

### Docs and agent guidance follow the code

`docs/{en-US,zh-CN}/DESIGN.md`'s `MathError` example, the two skill
cheatsheets, and a new rule in `.github/instructions/yo-design.instructions.md`
saying to reach for `derive(Error)` first and hand-write only when the render
is not one message per variant.

Editing two bundled skill files moves six cli-case tree goldens (`init*`,
`skills-install*`) — re-recorded with the in-repo binary, and the diff is only
the two files' hashes.

### Gate

| gate | result |
| --- | --- |
| `yo check ./std` with the **v0.2.29 seed** | 174/174 |
| Full suite (`./tests`, minus `internal` + `cli-cases`) | **3873 passed, 0 failed** |
[...]

PR #518, opened 2026-09-09
derive: a rule's own error reaches the user instead of a generic message

shd101wyy:
Writing a `derive_rule` was writing blind. A rule can validate its
`trait_params` and raise a precise `comptime_assert` naming the offending
variant — and none of it reached the user:

| where the rule raised | what the user saw |
| --- | --- |
| the result position | `derive: derive rule function failed` |
| a `::` binding in the body | **nothing** — `check` AND `compile` exit 0 |

The second row is the serious one: a `derive` that produced no impl reported
success. This came out of writing `derive(Error)` (#513), where it cost that
rule its keyed-message design — the pieces work in isolation but composing them
inside a rule gave no diagnosable error.

## Root cause

`evaluate_derive` called the rule through the **3-argument
`evaluate_expression`**, which is `_evaluate_expression_wrapper`
(`src/evaluator/exprs/_expr.yo`) — a handler that swallows every throw and
returns an err-expr. So `derive` never observed the exception; it observed a
missing expr-info value and manufactured a generic message from it.

The swallow is **not** one of the trial channels, which is what made it hard to
find: `YO_DEBUG_SWALLOW=1` prints no `[swallow]` line for it at all. Any
evaluator caller reaching for the 3-argument form silently loses errors.

## Fix

The machinery already existed in the same file. `_derive_eval_impl` — guarded
evaluator + module-level message box + caller re-raise, used for the generated
`impl(...)` — is renamed `_derive_eval_guarded` (its body was already generic)
and the rule call now goes through it. The rule's message is re-raised through
`derive`'s real `exn`, and the diagnostic anchors at the failing
`comptime_assert`, with the `derive(...)` statement as a secondary anchor:

```
error[E1101]: derive on "P": the derive rule failed: the rule's OWN diagnosis, …
   --> main.yo:21:12
   |
21 |   __why :: comptime_assert(false, "the rule's OWN diagnosis, …");
   |            ^^^^^^^^^^^^^^^
   --> main.yo:26:11
   |
26 | derive(P, MyRule());
```

Its doc now states the rule for future edits: **no derive-internal evaluation
may use the 3-argument `evaluate_expression`.** The two generic messages
survive only for a rule that returns *without* throwing and still leaves no
comptime value, and now say exactly that.

`_derive_strip_error_prefix` is factored out of the existing impl re-raise and
handles **both** prefix spellings. It only stripped `Error: `, while
`format_error_message` emits the lowercase `error: ` the renderer uses — so the
first cut of this change read `failed: error: …`.

## Pinned by a cli-case, not a unit test

`comptime_expect_error` only checks that *something* threw, never which
message. So `tests/cli-cases/check-derive-rule-error-reaches-the-user` runs
`yo check` on a rule that asserts and matches the rule's own text. Its
keep-match anchors on the `the derive rule failed: ` wrapper rather than the
message alone, because the diagnostic echoes the offending source line — which
*is* the `comptime_assert` call, so the message otherwise appears twice.

## The other half is split out, not fixed

`issues/anonymous-module-trial-swallows-a-top-level-derive.md`: a top-level
`derive` whose evaluation throws can still be eaten by the anonymous-module
begin-expr trial. That is right for an order-independent `::` binding and wrong
for `derive` — nothing re-runs a `derive` on a later pass, so a swallowed
failure is a permanently missing impl behind a green `check`. It needs a
lazy-bindings design decision and does not belong inside a one-call-site fix.
Reproducer 2 (`issues/repros/derive-rule-error-vanishes-entirely.yo`) still
exits 0, and the issue says so.

## Seed note

`std/` cannot lean on this until the seed ships it, so `derive(Error)`'s
declaration-order constraint stays for one more release; the keyed redesign it
unblocks can then land without touching a call site, since declaration order is
a valid keyed list.

## Gate

| gate | result |
| --- | --- |
| `yo check ./src` with the v0.2.29 seed | 271/271 |
| `tests/cli-cases/check-derive-rule-error-reaches-the-user` | 1/1 |
| Both repros re-run against the built binary | reproducer 1 now reports the rule's own message; reproducer 2 unchanged, as documented |
| Full suite, scorecard, fixpoint | running; posted below |


shd101wyy:
### Brought up to date with `develop` and re-gated — this run also covers the six-way std combination

`develop` gained #511, #512, #513, #514, #515, #516 and #517 while this sat, so
it was merged in (beaa68c4). Two conflicts, one of them silent:

- **Predicted:** #513 (now in `develop`) and this PR both ADD
  `issues/repros/derive-rule-error-{is-swallowed,vanishes-entirely}.yo`.
  Resolved by taking this PR's copies, which carry the updated headers.
- **Not flagged by git:** the issue doc ended up in *both* places —
  `issues/derive-swallows-the-rule-error.md` (from #513, marked **OPEN**) and
  `issues/fixed/derive-swallows-the-rule-error.md` (from this PR). From this
  side it was a rename while `develop`'s copy landed at the old path, so git
  saw no conflict. Two copies of one issue with contradictory status is the
  kind of thing that survives for months; the stale open copy is removed.

### A wording fix that was false, not merely stale

`std/error.yo` and the plan both said the keyed-message design "is not
diagnosable today". After this PR it **is** diagnosable — what actually holds
`derive(Error)`'s declaration-order constraint in place now is the **seed
gate**: `std/` cannot rely on the fixed compiler until the seed ships it. Both
sites now say that, so the constraint is not misattributed and nobody tries to
lift it a release early.

### Gate

| gate | result |
| --- | --- |
| `yo check ./src` with the v0.2.29 seed | 271/271 |
| Full suite on the merged tree | **3891 passed, 0 failed** — includes `tests/dyn.test.yo` and both `derive(Error)` tests |
| CLI golden scorecard | **PASS 90, GOLDEN-DIFF 0, NO-GOLDEN 0**, 1 network SKIP |
| `tests/cli-cases/check-derive-rule-error-reaches-the-user` | 1/1 |
| CI on the pre-merge head | 23 checks passed, 0 failed |

That suite run is doing double duty: #511/#513–#517 were admin-merged while
behind one another, so their *combination* was never CI-tested as a set. This
tree is `develop` + this PR, so 3891/0 is the evidence for the combination as
well as for the fix.

PR #519, opened 2026-09-09
plans: re-verify the P1 battery lists against the code — 20 rows were stale

shd101wyy:
The four P1 paragraphs (Encoding, I/O, Core, Concurrency) are a snapshot from
2026-09-06 and were still listing as open a great deal that has since landed —
so the list could not be used as a worklist without re-checking every row by
hand, which is what I did today. Checked against `develop` at `05510302b` and
split each paragraph into **LANDED** / **STILL OPEN**.

### Listed open, actually landed

`Url.join`/`query_pairs`/`path_segments`; `JsonValue` mutation,
`is_*`/`as_i64`/`as_u64`, `pointer`, integer arms; TOML values; `GlobPattern` +
`glob()`; `Stdout.write_string`; `Reader.read_exact`; lazy `read_dir`;
`Metadata.modified`; `SocketAddr`/`IpAddr` Eq/Hash/Ord/Clone + `parse`;
`TcpStream.local_addr`; `TcpListener.incoming`; `Error` `is(T)` (the free
`error_is`); `Context(msg, source)`; `Default` across the type set; `bench`
`black_box`; `log`'s `Sink` + `YO_LOG`; `Thread(T).join -> T`;
`Semaphore.with_permit`; `TryRecvError`.

The `checked_*` integer family and the `f64`/`f32` methods were the same kind
of staleness and were corrected in #516 and #515 earlier today.

### Genuinely still open, now stated as such

`Url.set_*`; `EncodingError` offsets; module-prefix stutter; a `Seek` trait
(only a bare `lseek` in `std/sys/seek.yo`); `OpenOptions`; a real
`SystemTime`/`UNIX_EPOCH`; `IpAddr.parse_v6`; `Child` handles as
`Reader`/`Writer`; `Watcher Dispose`; `UdpSocket.recv_from -> (n, from)` — it
still takes raw `*u8`/`*u32` out-params, a C-shaped signature sitting in std;
HTTP `StatusCode`/`HeaderMap`/byte bodies; a documented `downcast` and
`ErrorChain`; the unchecked integer forms plus `abs_diff`/`div_euclid`/
`midpoint`/`isqrt`/byte conversions; `Sender`/`Receiver`; waker-based
yield/channel/mutex; `Once.call`; `_raw_*` privatization; `JoinHandle Dispose`;
`interval`; `spawn_blocking`; a concurrent `Mutex` test.

### Two distinctions the old text did not make

- **`Mutex.try_lock` / `Condvar.wait_timeout` / `RwLock.try_*` are blocked on
  the SEED, not on design.** They need new `__yo_*` runtime macros, and the
  Linux half changes `__yo_cond_init`, which the GC's stop-the-world condvar
  shares. Compiler PR first, std PR one release later. Listing them beside
  ordinary API work invited someone to start and stall.
- **A live runtime bug sits under Concurrency:**
  `issues/yield-resumption-order-diverges-on-macos-ci.md`. Recorded in the
  paragraph so the group is not picked up as pure API work while a scheduling
  bug is open beneath it.

Docs-only, so it needs `--admin` to merge (the fast path skips the required
contexts).


shd101wyy:
Rebased onto `develop` (now at #518). No `plans/` conflict, and #518's wording fix — that `derive(Error)`'s declaration-order constraint is held by the SEED GATE rather than by diagnostics — carried through intact.

Spot-checked all fourteen "STILL OPEN" claims against `develop` rather than trusting the labels: `Seek`, `OpenOptions`, `SystemTime`, HTTP `StatusCode`/`HeaderMap`, `Url.set_*`, `IpAddr.parse_v6`, `ErrorChain`, `Sender`/`Receiver`, `JoinHandle Dispose`, `interval`, `spawn_blocking`, unchecked integer `abs`, `to_be_bytes` — every one genuinely absent. So the refresh is accurate in both directions, not just re-labelling the easy half.

PR #520, opened 2026-09-09
std/prelude: the remaining integer arithmetic — and two signed-overflow bugs the tests caught

shd101wyy:
Completes §4's Core row.

**In the `where(T <: Integer)` blanket:** `pow`/`saturating_pow`,
`checked_neg`, `wrapping_mul`/`overflowing_mul`, `div_euclid`/`rem_euclid`,
`midpoint`, `isqrt`.

**Behind a new `SignedInteger` marker:** `abs`, `signum`, `wrapping_neg`,
`wrapping_abs`. Rust omits `abs`/`signum` on unsigned types — one is the
identity, the other can only answer 0 or 1 — and a marker trait is how a
blanket gets restricted to half the integers, since `where(T <: Integer)`
covers all ten.

**Per-type:** `abs_diff` and the four byte conversions. `abs_diff` returns the
receiver's *unsigned partner* (`i32::abs_diff -> u32`) and the conversions
return an `Array(u8, N)` sized by its width; neither is nameable from a
blanket, the same reason the bit batteries are per-type.

`usize`/`isize` get `abs_diff` but deliberately **no** byte conversions: N
would be the target's pointer width, and a type-level size cannot be derived
the way `_USIZE_BITS` derives a value. Hard-coding 8 would be silently wrong on
the wasm legs, so it is not offered and the doc says to go through a
fixed-width type (`u64(n).to_be_bytes()`).

## Two bugs in this PR's own first draft

Both caught by tests written for the extremes, both the same class as the span
bug fixed in #515 earlier today — a signed intermediate that overflows, which
is **undefined behaviour** in C:

- **`midpoint`** was `lo + (hi - lo) / 2`, which looks overflow-free and is
  not: `i64.MAX - i64.MIN` does not fit. Now `(a & b) + ((a ^ b) >> 1)`, where
  no intermediate can leave the type — the AND carries the common bits, the XOR
  the differing ones, and half of those is the rest of the average. `>>` is
  arithmetic on signed in Yo, so it floors, making `i64.MIN.midpoint(i64.MAX)`
  equal `-1`.
- **`isqrt`** seeded Newton at `(self + 1) / 2`, which overflows at `T.MAX`.
  Now seeded at `self / 2`, with `self < 4 => 1` peeled off first: `self / 2`
  only dominates the root from 4 up, so 2 and 3 would otherwise stop early and
  answer 2.

On two's-complement hardware the wrapped value is usually the one you wanted,
which is why only assertions at MIN/MAX find these.

## Notes on two Yo constraints hit

- `overflowing_mul` returns `Tuple(T, bool)`, not `(T, bool)` — the sugar is
  rejected in a generic context ("cannot store a type value in tuple"). Matches
  the existing `overflowing_add`/`overflowing_sub`.
- The byte arrays use the `Array(u8, N)(...)` constructor rather than
  `fill` + indexed assignment: `Array.fill` is not resolvable this early in the
  prelude, and a direct construction needs no mutation anyway.

## Gate

| gate | result |
| --- | --- |
| `tests/int_checked_arithmetic.test.yo` | **28/28** |
| Full suite, scorecard, fixpoint | running; posted below |

The tests spell out the C-vs-Euclidean sign contrast explicitly (`(-7) / 2` is
`-3` and `(-7) % 2` is `-1`, against `div_euclid` `-4` and `rem_euclid` `1`),
check the `q*d + r == n` identity over all four sign combinations, and cover
`MAX.midpoint(MAX)`, `MIN.midpoint(MAX)`, `MAX.isqrt()` with the `(r+1)²> MAX`
bound checked by division, `i32.MAX.abs_diff(i32.MIN) == u32.MAX`, and
round-trips at every width including `i64.MIN`.


shd101wyy:
### A second defect came out of this branch's scorecard — fixed, not re-recorded

The first scorecard reported `compile-timeout` as **NO-GOLDEN — vacuous**
(`stdout_keep_match matched nothing`), which is a real bug by the project's own
rule, never recording residue. It was:

```
$ yo compile main.yo --compile-timeout-ms 1 --emit-c --skip-c-compiler
error: Array element at index 1 has incompatible type:
- Expected: u8
- Given   : Type(1)
     --> std/prelude.yo:5584:7
```

blaming this PR's new prelude code — which the controls exonerate:

```
$ yo compile main.yo --emit-c --skip-c-compiler                            # rc=0, silent
$ yo compile main.yo --compile-timeout-ms 5000 --emit-c --skip-c-compiler  # rc=0
```

Clean with no deadline and with a generous one; only the 1 ms budget produces
it. The evaluator deadline is LATCHED so a swallowing handler cannot outrun it —
but that only wins if the next event **is a dispatch**, and a half-evaluated
expression can format its own diagnostic without dispatching. The race is with
error *formatting*, not with evaluation.

Fixed in 6ee4277 at the two REPORTING EDGES: `_expr.yo` exposes
`evaluator_deadline_tripped()` / `evaluator_deadline_message()`, and both
`mm_preload_prelude` and `compile`'s entry-eval failure path consult it first.
After a trip the evaluator's state is unreliable, so the timeout is not merely
the more useful message — it is the only one still true.
`issues/fixed/a-tripped-deadline-is-masked-by-the-next-type-error.md` has the
detail. `compile-timeout` now passes with **no golden re-record**.

That is the second time prelude growth has exposed a different defect at these
same edges today (the first was #516's double-report). The case catches both
only because its keep-match asserts THE failure rather than A failure.

### The four re-recorded goldens are accounted for, not just re-baselined

- `check-watch-once`, `check-forward-ref-async-body`: `parsed 1203` → **1221**,
  a delta of exactly the 18 top-level expressions added (`SignedInteger` + its
  export + 5 marker impls + the signed blanket, plus 10 per-type impls).
- `lsp-member-definition`: a prelude definition at line 7394 → **7998**, a
  shift of exactly the 604 lines inserted above it.
- `lsp-analysis-resilience`: `Content-Length` plus one semantic change. I
  diffed the completion **label sets** rather than eyeballing it — nothing
  removed, exactly one added: `SignedInteger`, the new marker trait appearing
  where it should.

### Gate

| gate | result |
| --- | --- |
| `yo check ./std` / `./src` with the v0.2.29 seed | 173/173, 271/271 |
| `tests/int_checked_arithmetic.test.yo` | 28/28 |
| Full suite | **3899 passed, 0 failed** |
| CLI golden scorecard | **PASS 90, GOLDEN-DIFF 0, NO-GOLDEN 0**, 1 network SKIP |
| Fixpoint | running |

shd101wyy:
Bootstrap fixpoint: **FIXPOINT_HOLDS** — `STAGE2_RC=0`, `STAGE3_RC=0`, `CLANG_RC=0`, `stage2 hollow=0`.

Local gate complete: seed `check ./std` 173/173 and `./src` 271/271, suite **3899/0**, scorecard **90 PASS / 0 GOLDEN-DIFF / 0 NO-GOLDEN**, fixpoint holds.

PR #522, opened 2026-09-09
std: Seek trait, OpenOptions, SystemTime — plus documented downcast and a diagnostic that cited a nonexistent is()

shd101wyy:
Closes the I/O row's `Seek`/`OpenOptions`/`SystemTime` group and the
"documented `downcast`" row from `plans/STD_API_STABILIZATION.md`.

## `SystemTime` + `UNIX_EPOCH` (`std/time/instant.yo`)

A second clock, deliberately **not** interconvertible with `Instant`: `Instant`
reads `CLOCK_MONOTONIC` (no epoch, never steps, only differences mean
anything), `SystemTime` reads `CLOCK_REALTIME` (anchored at `UNIX_EPOCH`,
steppable by NTP or by hand). `duration_since`/`elapsed` therefore return
`Result(Duration, SystemTimeError)` — Rust's shape — because a backwards step
is a real outcome, not something to hide behind a `Duration.zero()`.

All of its arithmetic stays on the `(secs, nanos)` **pair**. Converting either
side to a single `i64` nanosecond count overflows past year 2262, and signed
overflow is UB in C. Every constructor normalizes `nanos` into `[0, 1e9)` with
a **floor** second, so one instant has exactly one representation: without
that, `(-1, -4e8)` and `(-2, 6e8)` are the same timestamp and compare unequal
under the field-wise `Eq`/`Ord`, and `as_unix_secs` rounds toward the epoch
instead of down.

## `Metadata.modified` / `accessed` / `status_changed` (`std/fs/metadata.yo`)

Now return `SystemTime` built from `statx`'s seconds **and** nanoseconds. The
seconds-only `*_time` accessors were discarding `mtime_nsec` outright; they are
kept for callers that want the raw field, with a test pinning
`modified_time() == modified().as_unix_secs()`.

## `Seek` (`std/io/index.yo`), implemented by `File`

Synchronous while `Reader`/`Writer` are async, because moving a position is
arithmetic on a handle's own state: `File`'s reads and writes are positional
(`pread`/`pwrite`), so its descriptor sits at offset 0 forever and `File.seek`
never touches it. Parameterized over the reference-point type (`Seek(From)`)
rather than re-declaring `SeekFrom`, so there is one spelling of "from where"
in the tree.

**No `rewind`, unlike Rust.** Every absolute move needs a `From` *value* naming
the beginning and a trait generic over `From` cannot name one, so a
`rewind(self, from, exn)` default would make the caller pass the very thing
`rewind` exists to hide. The convenience is inherent on `File` instead.

## `OpenOptions` (`std/fs/types.yo`)

A plain value struct with functional setters rather than the
`ref(struct(...))`-with-mutation shape `Command` uses: a flag bag needs no
allocation or refcount, and an immutable builder lets one base be reused for
several opens. It exists because `OpenMode`'s five variants cannot express
read+append, create-without-truncate, or read+write+create, and growing that
enum combinatorially is the wrong answer.

`to_flags()` **rejects** the four combinations POSIX does not diagnose — no
access mode at all, `truncate` without write access, `truncate` with `append`,
`create` without write access — because `open(2)` quietly ignores whichever
flag does not apply and hands back a file that behaves differently from what
was asked. `File.open_opts` raises a rejection as a `Context` over
`IoError.InvalidInput`, so `to_string()` names the contradiction while
`source()` still reports the kind Rust would report. Note the access mode is a
**value**, not a bit set (`O_RDONLY` is 0), so it is chosen rather than OR-ed.

## Documented `downcast` (`docs/{en-US,zh-CN}/DYN_DESIGN.md`)

It was the one way to recover a concrete type from a `Dyn`, mentioned in
passing by a single DESIGN.md snippet and nowhere explained. The new section
covers the `Option(T)` result and why there is no unchecked cast, the single
pointer compare against the vtable's `__yo_type_id` (whose canonical value is a
per-type static's **address** — no string compare, no RTTI table), the
owned/RC'd result, the box-unwrapping for a value/newtype target, and the
statically-`.None` lowering for a target no `dyn()` in the program ever wraps.

## Bugs fixed along the way

**`typeid`'s diagnostic cited an `is()` that does not exist.** It read
`Use is() or downcast() for runtime type checks on Dyn values`, and there is no
`is()` in the language — no `BF_IS` in `src/expr.yo`'s builtin table and no
definition anywhere in `src/` or `std/`; grepping the tree finds only that
message. Following the advice produced a second, unrelated
`Variable "is" not found`. It now names only APIs that exist. Write-up in
`issues/fixed/typeid-diagnostic-suggests-a-nonexistent-is-builtin.md`, repro in
`issues/repros/`, and the write-up records a second gap it uncovered: wrapping
the call in `comptime_print(...)` replaces the inner diagnostic with the
wrapper's own "Failed to evaluate argument" and the real message is lost.

**`derive(Error)` needs `ToString` in the *calling* file's scope** — documented
in `std/error.yo` after hitting it while adding `SystemTimeError`, where it
cost a `Variable "ToString" not found` pointing into `std/error.yo` rather than
at the call site.

## Coverage

- `tests/time/instant.test.yo` — **14/14** (9 new `SystemTime` tests: epoch
  timestamp, backwards `duration_since`, borrow across the second boundary,
  add/sub round-trip, pre-epoch floor normalization, `elapsed`, `Eq`/`Ord`, and
  a year-5138 timestamp that would overflow a nanosecond count).
- `tests/fs/metadata.test.yo` — **10/10** (2 new: mtime near now with the
  seconds-only accessors pinned against the `SystemTime` ones; mtime never
  going backwards across a rewrite).
- `tests/fs/file.test.yo` — **29/29** (10 new: access mode chosen not OR-ed,
  each flag mapped, each rejected combination, setter non-mutation,
  read+append, create-without-truncate, `create_new` on an existing path, the
  `Context` message, `rewind`, and a generic function bounded on
  `Seek(SeekFrom)` — the last being the only thing that proves the impl is
  registered rather than the inherent method satisfying the call).
- `tests/cli-cases/check-typeid-on-a-value` — pins the new diagnostic text.
  A `comptime_expect_error` cannot cover it: that asserts only *that* an
  expression fails, never with which message, and the message is the whole
  defect.

## Local gate (all green)

| gate | result |
| --- | --- |
| `yo build` with `YO_STD=<worktree>/std` | rc=0 |
| full suite (`tests`, minus `internal`/`cli-cases`) | **3920 passed, 0 failed** (incl. the `tests/dyn` canary) |
| CLI golden scorecard | PASS **91**, GOLDEN-DIFF 0, NO-GOLDEN 0, SKIP 1 |
| `gates_fast.sh` (7 gates) | rc=0 — `check ./std` **173/173**, `check ./src` **271/271**, corpus PASS 156 |
| `fixpoint_only.sh` | **FIXPOINT_HOLDS**, stage2 hollow=0 |


shd101wyy:
Rebased onto `develop` after #521 landed. The only conflict was in `plans/STD_API_STABILIZATION.md`, where both PRs appended to the same paragraph — resolved by keeping BOTH hunks (#521's `ErrorChain` blocker diagnosis and this PR's `downcast` documentation note), in landing order.

No `.yo` file changed in the rebase, so the local gate reported above still stands on the same code.

Also converted the three import lines this branch ADDS from `open(...)` to named imports, since `open(...)` is being removed from the language. Pre-existing `open(...)` lines are left to that sweep. `check ./std` 173/173 and the three affected test files re-run green (metadata 10/10, file 29/29, instant 14/14).

PR #534, opened 2026-09-09
verifier: V3 match/datatype encoding — value enums verify as SMT datatypes

shd101wyy:
## Summary

Closes the last unchecked row of V3's subset table: value enums verify as SMT datatypes. **Validated end-to-end locally against z3 5.1.0**: the fixture's obligation encodes to a well-formed `declare-datatypes` script and **proves (unsat)**.

- **terms** (`src/verifier/terms.yo`): `VcSort.DataS(name)`; `VcTerm.Ctor/Test/Proj` — constructor application, `(is-<Ctor> t)` tester, accessor projection; `VcQuery.datatypes` carries the combined `declare-datatypes` block (rendered once, before the constants).
- **vc** (`src/verifier/vc.yo`): `_type_sort` maps value enums to `DataS`; `_note_enum_datatype` registers `<dt>|<ctor> ((<ctor>_<label> <sort>)...)` entries idempotently per context — accessors are mangled `<ctor>_<label>` because they share the datatype's whole SMT namespace (two variants with an `n` field must not collide); construction dispatch for `.Variant(...)` with value args; the match walk folds arms into an ite tree over `Test`/`Proj` with pattern bindings projected off the scrutinee through the mangled accessors (bindings saved/restored around arm bodies); walkers extended for the three new term kinds.
- **encode** (`src/verifier/encode.yo`): `declare-datatypes` rendered first, each constructor individually parenthesized per SMT-LIB 2.6 with the field group flattened (`_wrap_ctor_segments`); `(is-C t)` / accessor projections; `_sorted_lines` guards its `ptr().unwrap()` for empty lists.
- **tests**: `tests/internal/verifier_match.test.yo` — a real-solver proving battery (gated `YO_TEST_Z3=1`, the existing verifier-test convention) plus solver-free encoding pins; fixture `tests/spec/fixtures/valid/match_enum.yo` — `bind_floor` keeps the ensures provable with unconstrained inputs via `(v - v)` wrap-zero, `toggle` exercises a construction inside a match arm body under `:=` + `cond`.
- **docs**: en+zh `FORMAL_VERIFICATION.md` subset tables updated (match over value enums ✅, structs/tuples still in progress) with the datatype-encoding paragraph; plan banner records the row landing.
- **issue housekeeping**: `issues/match-binding-shadows-scrutinee-…` moves to `issues/fixed/` (fixed by #527, which this builds on); the walk's own `.Test(_, child)` bindings deliberately avoid the `t` name so local runs keep working under the v0.2.29 seed.

## History

Originally stacked on #527 as #528 (auto-closed when #527's branch was deleted); #533 carried it but GitHub stopped creating workflow runs for that branch after a force-push raced a cancel (four pushes, zero runs — verified via the Actions API while other PRs triggered normally). This PR is the identical commit chain on a fresh branch.


shd101wyy:
Superseding: this PR's opened event landed during the Actions run-creation stall and no checks ever attached; subsequent synchronize pushes produced zero runs while other branches triggered normally. Replacement PR carries the identical commits.

PR #550, opened 2026-09-10
std: OrderedMap position index (swap_remove, index_of, get_index); async race_first/any_first

shd101wyy:
Two small P1 items from `plans/STD_API_STABILIZATION.md`, plus the plan
corrections that writing them forced.

## `OrderedMap` gains a position index

`IndexMap::swap_remove` is O(1) because `IndexMap` keeps a key→position map.
`OrderedMap` kept only `HashMap(K, V)` plus an `ArrayList(K)` order list, so
finding the slot to swap into meant scanning the order list — which would have
made `swap_remove` O(n) and pointless, since avoiding the walk is the whole
reason it exists beside `remove`.

`_index : HashMap(K, usize)` is maintained at the four sites that write
`_order` (`new`, `insert`, `remove`, `clear`), and it pays for itself three
more times:

- `remove` (Rust's `shift_remove`) used to rebuild the entire order list into a
  fresh `ArrayList` to skip one key. It now asks the index where the key is,
  calls `ArrayList.remove` on that slot, and re-records only the TAIL — same
  O(n) worst case, no allocation, one pass instead of two, and O(1) when the
  removed key is near the end.
- `index_of` / `get_index` (Rust's `get_index_of` / `get_index`) fall out of it
  — the positional half of `IndexMap`'s surface `OrderedMap` could not offer
  before.

Both removals PANIC when a key is in `_map` but missing from `_index` rather
than defaulting the position to 0: a default corrupts the order list silently
on the next operation, and the invariant is internal, so a violation is a bug
in this file rather than something a caller can cause.

## `race_first` / `any_first`

`race` and `any` hand back an INDEX and leave a manual contract behind: every
handle must still be awaited exactly once, and the losers aborted first if
their results are unwanted. A caller who forgets it leaks every loser's state
machine — the finding the plan's "`JoinHandle` `Dispose`" row was really about.

That prescription is wrong twice over. `JoinHandle(T)` is a bare copyable
`struct(__future : *(T))`, not an `Rc`, and `Dispose` is `where(Self <: Rc)`.
And abort-on-drop is not the semantics Yo's async `JoinHandle` has: dropping it
DETACHES, exactly as dropping Tokio's does, which is what makes
fire-and-forget `io.spawn` work at all. Tokio spells the other behaviour
`AbortOnDropHandle` and makes it opt-in for the same reason.

So the leak is fixed in the combinators: `race_first`/`any_first` abort AND
await every loser (aborting alone marks the task and leaves its await
outstanding — the same leak one step later) and return the winner's value.
`race`/`any` keep their index shapes.

## Plan corrections measured against the code

- the two contradictory `JoinHandle Dispose` rows
- `async/channel.try_recv` already returns `Result(T, TryRecvError)` over the
  same error type (since #506), not `Option(T)`

## Coverage

- `tests/collections/ordered_map.test.yo` → 20/20
- `tests/async/combinators.test.yo` → 12/12
- CI carries the rest of the battery.

### DIFF
--- a/plans/STD_API_STABILIZATION.md
+++ b/plans/STD_API_STABILIZATION.md
@@ -413,7 +413,7 @@ window as D9–D18):
 | row as written | actual state |
 | --- | --- |
 | `Alignment` exported | done — `std/fmt/writer.yo:22` |
-| `JoinHandle` `Dispose` | done — Yo's `Thread` IS Rust's `JoinHandle` and has `Dispose` (`std/thread.yo:88`, detach-on-drop) |
+| `JoinHandle` `Dispose` | **that row conflated two types, and the ask is answered differently — see the `race_first`/`any_first` note below.** Yo's `Thread` is Rust's *thread* `JoinHandle` and does have `Dispose` (`std/thread.yo:88`, detach-on-drop); Yo's async `JoinHandle(T)` is TOKIO's, and cannot carry `Dispose` at all |
 | HTTP byte bodies (*"`parse_response` string-concats the body — binary responses are broken client-side"*) | **done, and the claim is stale**: the body is copied byte-wise and wrapped with unchecked `String.from_bytes` (`std/http/http.yo:363-374`), byte-transparent end to end, pinned by `tests/http/server.test.yo` |
 | `OrderedMap` `IntoIterator` | done via D14 — `std/collections/ordered_map.yo:319` |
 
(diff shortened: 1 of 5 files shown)


shd101wyy:
Rebased onto `develop` (now carrying #562 and #561) and verified locally with the freshly built compiler: `yo check ./std --std-path ./std` 174/174, `tests/async/combinators.test.yo` 12/12, `tests/collections/ordered_map.test.yo` 20/20. The plan-doc conflict against the waker record was resolved keeping both records in landing order.

PR #554, opened 2026-09-10
std/encoding/toml: a real TOML v1.0.0 parser and serializer

shd101wyy:
Closes the last Encoding gap in `plans/STD_API_STABILIZATION.md` ("TOML
floats/arrays/dates/inline tables/escapes/comments/serializer") and the open
corruption issue that measuring it turned up.

## What was there

A line-based subset parser: `input.split("\n")`, one `cond` per trimmed line.
Besides missing every value form on the row, it returned `.Ok` with a
**different document** for twelve inputs, undetectably
(`issues/fixed/toml-parse-returns-ok-with-corrupted-data.md`): `[[fruit]]`
became a table keyed `[fruit]`, `[server.http]` one flat root key, `"my key"` a
key with its quotes still on, a duplicate key two entries `get` could not
reach, `s = "a" # "c"` the string `a" # "c`, and
`n = 99999999999999999999` a wrapped `Int`.

Nothing on the row's list was already implemented: `3.14`, `[1, 2]`,
`1979-05-27`, `{ a = 1 }` and `\n` all returned `.Err` or the wrong value.

## What landed

A byte scanner over the whole document, in the shape `csv.yo` and `json.yo`
already use. Accepted: comments anywhere a value or key can be followed by one;
bare, quoted and dotted keys; basic and literal strings, single and multi-line,
with `\b \t \n \f \r \" \\ \uXXXX \UXXXXXXXX`, the dropped opening newline and
the line-ending backslash; decimal/hex/octal/binary integers with underscores;
floats including `inf`, `-inf`, `nan` and exponents; all four date-time forms
with fractional seconds and offsets; arrays (heterogeneous, nested, trailing
comma, newlines and comments inside the brackets); inline tables (nested,
dotted keys, no trailing comma, no newline); `[table]` and
`[[array of tables]]`. Everything else is an `.Err` with the byte offset of the
fault.

Plus a serializer — `toml_stringify`, and `ToString` — that round-trips
everything the parser accepts, with deterministic output.

## Shape decisions

| decision | why |
| --- | --- |
| `TomlError`, not `EncodingError` | that enum is the shared error of the BYTE codecs (hex/base64/utf16); ten document-level variants there would make every `hex` caller match on them. `JsonError` is the precedent, and D1 bans the `Result(_, String)` this module had. All ten variants carry an offset, so `pos()` is total |
| `TomlDateTime` = `std/time`'s `DateTime` + a four-way `kind` | the tag is unavoidable (`DateTime` has no "no offset" state and no date-less/time-less form), but the components and `DateTime.new`'s leap-aware validation are reused rather than duplicating the leap-year table. Its `Eq` is fieldwise, unlike `DateTime`'s instant `Eq`: `07:32:00Z` and `08:32:00+01:00` are different TOML values |
| a `(path, kind)` provenance table | TOML's redefinition rules are about HOW a table was created, not whether the path was seen. Paths are length-prefixed so `["a.b"]` and `["a", "b"]` cannot collide; array elements append `@<index>;` so `[[fruit]] … [fruit.variety]` attaches to the fruit most recently opened |
| serializer writes scalars before headers | a `[header]` changes which table the following bare keys belong to, so the other order would read back into the wrong table. Insertion order within each pass ⇒ the text is a deterministic function of the tree (stringify-twice is asserted byte-identical) |
| shortest round-tripping float | `%g`'s six digits would turn `3.141592653589793` into `3.14159`; `%.17g` would turn `3.14` into `3.1400000000000001`. `.0` is appended when the rendering has no `.` or exponent, or TOML reads it back as an integer |
| `\u{XXXX}` deliberately rejected | Rust's spelling, not TOML's — accepting it would accept documents the `toml` crate rejects |
| `insert -> Option(TomlValue)` | D2's map-insert shape, matching `JsonValue.insert`. The old one returned `unit` and set every matching key |
| `## Stability unstable` for one release | the error type, three new variants and `insert` are breaking, and documents that used to "parse" now `.Err`. Nothing in the tree consumed the module, so the blast radius is external users — release-notes material |

## Tests

`tests/toml/toml.test.yo` → `tests/encoding/toml.test.yo` (its eight siblings
all live there; it was the last one outside), 10 happy-path cases → **36
tests**: one assertion per corruption row *with its byte offset*, a round trip
per value kind, the malformed inputs that must be rejected, and `nan`/`inf`
asserted with `is_nan()`/`is_infinite()` since `nan != nan`.

```
yo check ./std --std-path ./std              # 173/173
yo test ./tests/encoding --std-path ./std --parallel 1   # 215 passed (whole directory)
yo fmt --check .                             # clean
```

## One compiler bug found, filed not fixed

`issues/unicode-escape-accepts-non-hex-digits.md` (+ reproducer): `"\uZZZZ"`
compiles to a NUL byte and `"\u{41}"` to U+0410 rather than U+0041, because
`_hex_digit_val` (`src/evaluator/values/string.yo`) returns 0 for a non-hex
rune instead of failing; and the backtick template scanner has no `\u` entry at
all, so a well-formed escape in a template (`` `x\u0041y` ``) stays eight
literal bytes where `"x\u0041y"` is `xAy` — one spelling, two meanings,
no diagnostic either way. The measured escape
tables of both literal forms are now recorded in the syntax cheatsheet and
instructions, which is why the two `skills-install` CLI goldens are re-recorded
(their `expected_tree` hashes every installed skill file).


shd101wyy:
Rebased onto `develop` and verified locally: `yo check ./std --std-path ./std` 174/174, `tests/encoding/toml.test.yo` 36/36 (`tests/toml/` moved into `tests/encoding/` in this PR).

PR #555, opened 2026-09-10
std/async: a `Stream` trait — async iteration, and `TcpListener.incoming`

shd101wyy:
Implements `plans/reference/ASYNC_ITERATION_STREAM.md` (moved out of `plans/backlog/` with a closing banner). No compiler change.

Yo had `Iterator` and it had `Future`, but nothing that was both, so every std API producing a sequence asynchronously invented its own shape — `accept` in the caller's own loop, `Watcher.next` answering `Option(FsEvent)`, `Channel.recv` plus a separate `try_recv` — and none of them composed. `std/async/stream.yo` is the missing abstraction:

```rust
Stream :: trait(
  Item : Type,
  next : (fn(self : Self, io : Io) -> Impl(Future(Option(Self.Item), Io)))
);
```

## What landed

Three implementors, all of which already had the shape by hand:

| type | its stream |
| --- | --- |
| `Watcher` (`std/fs/watch.yo`) | `Item = FsEvent` — its hand-rolled `next` BECAME the trait method, byte for byte |
| `TcpListener.incoming()` (`std/net/tcp.yo`) | `Item = Result(TcpStream, NetError)` — the deferred `STD_API_STABILIZATION` row, now closed |
| `Channel(T)` (`std/async/channel.yo`) | `Item = T` — `next` IS `recv` |

Lazy combinators `map` / `filter` / `filter_map` / `take` / `skip`, consumers `for_each` / `collect`, written once over `where(S <: Stream)` the way the `Iterator` combinators in `std/prelude.yo` are.

## Shape decisions, and why

- **`next`'s future is `Future(_, Io)`, not `Future(_, IoExn)`** as the plan doc wrote. A stream does not throw: a fallible one carries the failure IN the item (`Item = Result(T, E)`), so one bad item does not end the sequence and a consumer needs no handler. `Watcher.next` already used `Io`, and `Io` is what compiles.
- **`incoming`'s error is `NetError`, not `IoError`** as the plan wrote — it is the type `accept` THROWS, so a server loop moving from `accept` to `incoming` keeps one error vocabulary (`NetError.Io(IoError)` still carries the raw errno). The stream calls the raw `IO_tcp.accept` rather than the throwing method, because an `Exception` handler cannot be installed across a suspension.
- **`self : Self`, not `inout(self)`** as `Iterator` has: the future outlives the call and an `inout` borrow cannot cross a suspension, so every source is a reference-semantics type.
- **`.None` is terminal and stays terminal**; every combinator preserves it (`take` stops touching upstream once spent, `skip`/`filter` adopt upstream's end, `incoming` finishes when the listener closes — including from another task).
- The associated type is `Item`, carrying `DoubleEndedIterator`'s uncheckable coherence rule in its doc (the registry is keyed by (type id, label) with no trait discrimination). No std type implements both `Stream` and `Iterator`.

## What did NOT land, and why that is not a shortcut

- **`for_await` was written and then removed.** An `io.await` reached only through a MACRO EXPANSION is not counted as a suspension point, so the enclosing `io.async` body is emitted as a plain closure with a BLOCKING await — correct from `main`, a **deadlock** inside a spawned task (measured: a `for_await` over `incoming()` in a task hangs at rc=124 with no output). Shipping "correct from main, hangs in a task" for a server loop was not acceptable. `for_each`, `take` and the hand-written `while` + `io.await(s.next(io), io)` + `match` cover the ground.
- **An async `BufReader.lines` is blocked one level down.** A `Stream` reports failure in its item and `read_line` THROWS; a `ctl` handler can neither be installed across a suspension nor stored in a `ref` struct. The blocker is `Reader`'s error style, not the stream protocol.

Filed with minimal reproducers: `issues/io-await-inside-a-macro-expansion-is-emitted-as-a-blocking-await.md`, `issues/closure-argument-inside-an-io-async-body-loses-the-future-result-type.md`. Planned: `plans/backlog/FOR_AWAIT_NEEDS_MACRO_AWARE_ASYNC_TRANSFORM.md`, `plans/backlog/ASYNC_LINES_NEEDS_A_NONTHROWING_READ.md`, `plans/backlog/ASSOC_TYPE_BINDING_IN_FREE_FN_WHERE.md` (a free function's `where(T <: Trait(Assoc := A))` with a generic `A` binds nothing — true of `Iterator` too, so pre-existing).

## Verification

| gate | result |
| --- | --- |
| `yo check ./std --std-path ./std` | 174/174 |
| `yo fmt --check ./src ./std ./tests` | clean |
| `yo test ./tests/async/combinators.test.yo` | 27 passed |
| `yo test ./tests/async/channel.test.yo` | 8 passed |
| `yo test ./tests/net/tcp.test.yo` | 21 passed |
| `yo test ./tests/fs/watch.test.yo` | 6 passed |
| `yo test ./tests/io/bufio.test.yo` | 13 passed |
| `yo test ./tests/io/async_traits.test.yo` | 9 passed |
| `cli-diff-test.sh` skills-install, skills-install-zh, init, init-cwd, init-existing, init-build-test | PASS 6 |

The new tests cover every cardinality (empty, one, several), `.None` terminal AND staying terminal, each combinator (including the degenerate `take(0)`, `filter` that keeps nothing, `skip` past the end), a chain, a generic consumer, and each combinator awaited from inside a **spawned task** as well as from `main` — each bounded by `std/async`'s `timeout`, so a regression fails instead of burning a CI job. **Hollow-green probe done:** flipping one assertion to a false value turns the file red, so the batch really runs.

Six `tests/cli-cases/*/expected_tree` goldens are re-recorded because one `.github/skills/` file was edited (the hashes are `sha256` of the source); every affected case was re-scored green rather than assumed.

Docs: `docs/en-US/ASYNC_AWAIT.md` and `docs/zh-CN/ASYNC_AWAIT.md` gain an "async iteration" section; `.github/instructions/yo-design.instructions.md` and `.github/skills/yo-async-effects/async-effects-recipes.md` record the trait plus the three async-body traps measured here.

### DIFF
--- a/.github/instructions/yo-design.instructions.md
+++ b/.github/instructions/yo-design.instructions.md
@@ -222,6 +222,43 @@ my_fn :: (fn(io : Io) -> Impl(Future(Result(i32, IoError), Io)))(
 );
 ```
 
+## Async iteration — the `Stream` trait (`std/async/stream.yo`)
+
+The async analogue of `Iterator`, landed 2026-09-11
+(`plans/reference/ASYNC_ITERATION_STREAM.md`):
+
+```rust
+Stream :: trait(
+  Item : Type,
+  next : (fn(self : Self, io : Io) -> Impl(Future(Option(Self.Item), Io)))
+);
+```
+
+Four rules to keep when extending it or implementing it on a new type:
+
+- **The future's bundle is `Io`, never `IoExn` — a stream does not throw.** A
+  fallible stream puts the failure IN the item (`Item = Result(T, E)`, e.g.
+  `TcpListener.incoming`'s `Result(TcpStream, NetError)`), so one bad item
+  does not end the sequence and no consumer installs a handler. The knock-on:
+  an API that THROWS (`Reader.read`, `BufReader.read_line`) cannot become a
+  stream until it can return its failure —
+  `plans/backlog/ASYNC_LINES_NEEDS_A_NONTHROWING_READ.md`.
+- **`next` takes `self : Self`, not `inout(self)`** (`Iterator` uses `inout`):
+  the future outlives the call and an `inout` borrow cannot cross a
+  suspension. So every source is a `ref(struct(...))`.
+- **`.None` is terminal and stays terminal.** Every combinator preserves it.
+- **The associated type is named `Item`**, with the same uncheckable coherence
+  rule `DoubleEndedIterator` documents: the assoc-type registry is keyed by
+  (type id, label) with no trait discrimination, so a type implementing both
+  `Stream` and `Iterator` must give them the SAME `Item`. (`FromIterator` chose
+  `Elem` to dodge this; here no std type implements both.)
+
+There is deliberately **no `for_await` macro**: an `io.await` reached only
+through a macro expansion is compiled as a BLOCKING await and deadlocks inside
+a task (`plans/backlog/FOR_AWAIT_NEEDS_MACRO_AWARE_ASYNC_TRANSFORM.md`). The
+loop is `for_each`, or a hand-written `while` + `io.await(s.next(io), io)` +
+`match`.
+
 ## JoinHandle(T) — spawned task handle
 
 `JoinHandle(T)` is a builtin generic type returned by `io.spawn`. It wraps a pointer to the spawned future and allows awaiting its result.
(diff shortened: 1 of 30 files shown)


shd101wyy:
Rebased onto `develop` (now carrying #550's `race_first`/`any_first`, whose tests live in the same file). The `tests/async/combinators.test.yo` conflict was resolved keeping BOTH families — the file's module doc now names both, and the combined file is 31/31 locally. Also green with the compiler built from the new develop: `yo check ./std --std-path ./std` 175/175, `tests/async/channel.test.yo` 18/18, `tests/fs/watch.test.yo` 6/6, `tests/net/tcp.test.yo` 21/21, `yo fmt --check` clean.

PR #556, opened 2026-09-10
std/http: a connection-pooling HttpClient, and RFC 9112 §6.3 body-less framing

shd101wyy:
The framing half of HTTP keep-alive landed in #547; this is the client that uses it.

`HttpClient` **owns** its idle connections rather than being a module-global, because nothing in a free function could say when those sockets close — which is also why the free `fetch`/`fetch_with` keep their one-shot behaviour instead of gaining a hidden pool behind them.

```rust
{ HttpClient } :: import("std/http");

client := HttpClient.new();
a := io.await(client.fetch(`http://127.0.0.1:8080/one`, io), e);
b := io.await(client.fetch(`http://127.0.0.1:8080/two`, io), e); // same socket
client.close_idle();                                            // both closed
```

## One code path, parameterized by `pool : Option(HttpClient)`

The pooled and one-shot exchanges differ in three places — whether a connection is taken from a pool, whether `Connection: close` is sent, and whether the connection is returned or closed — and are identical in URL parsing, request building, framing, redirects and the deadline race. `_fetch_once`, `_fetch_follow` and `_fetch_deadline` each take the option, so there is exactly one copy of each. A second copy is where those would have drifted apart.

## Shape decisions

- **The pool key is `scheme://host:port`**, built from the URL's own host TEXT and compared, never inferred. Not the resolved address: two names that resolve to one address are still two peers as far as TLS, `Host` and virtual hosting go, and handing an `https://a` connection to a request for `https://b` speaks to the wrong peer under the first peer's certificate.
- **The idle cap is TOTAL** (default 16), because the resource being protected is the descriptor table and a per-host cap does not bound it — a client fanning out over a thousand hosts would hold a thousand sockets with a per-host cap of 1. When the pool is full the connection in hand is closed rather than evicting one already pooled (Go's answer), so a burst cannot churn the pool. `max_idle == 0` means never reuse.
- **The idle age limit** (default 30 s) is purged on the next request rather than by a timer — the only moment a `std` client is guaranteed to be running is a request. The real ceiling is the server's own timeout, 75 s on nginx and 5 s on Apache, so no value is right; 30 s discards what a short-timeout server has probably already dropped and the retry covers the rest.
- **Reuse follows RFC 9112 §9.3 in the RFC's order**: a server `Connection: close` ends it at any version; `HTTP/1.0` (and any unknown version) is close-by-default and needs an explicit `keep-alive`; a `Connection: close` we sent ends it; a response whose length is not determinable is delimited BY THE CLOSE and can never be followed by another message; and a 1xx retires the connection too, because the final response is still coming. `Connection` is a `#token` list, so the check is token-wise over every field line rather than a substring test. Each rule is its own test.
- **A connection the server closed while idle is retried once, on a fresh connection.** This is the behaviour a keep-alive client lives or dies by: the close usually reaches us only when we use the connection (the write buffers, the read that follows returns EOF with no response byte), and reporting that as a failure turns a pool into intermittent request failures. The trigger is exactly "came from the pool AND zero bytes arrived" — a freshly-opened connection is never retried, because there the EOF is real. The retry is limited to the methods RFC 9110 §9.2.2 calls idempotent: "the request was never processed" is not provable from this side, only "no response arrived", so a `POST` the server acted on before closing would run twice. Go replays those (its bodies are rewindable, as ours always are); this client reports `ConnectionFailed` and leaves the decision to the caller.
- **TLS pools identically**, through a `_Transport` enum over the two stream types with `Reader` implemented by a two-arm dispatch — the exchange, the pool and the retry are written once against that. It needed a TLS connection to be closable from plain code, and `TlsStream` had no `Dispose` at all.

## The framing layer learned RFC 9112 §6.3 rules 1–2

`read_http_message*` now takes a `MessageKind` (`Request` / `Response` / `HeadResponse`) instead of an `is_request` bool, and a 1xx, 204 or 304 response — or any response to a HEAD — ends at its header section whatever its framing headers claim. On a close-delimited connection that was harmless; on a reused one it is a **hang**, because those responses routinely carry no `Content-Length` and a keep-alive server never closes. The status code is read off the status line; HEAD cannot be, which is why the kind is a parameter and not a bool.

## Tests

`tests/http/http.test.yo` gains 15 cases (51 pass, up from 36) driving a raw `TcpListener` keep-alive server: `HttpServer` adds `Connection: close` to every response and so cannot demonstrate reuse, and a hand-written server is the only way to produce the four §9.3 response shapes. Each server accepts exactly N connections and closes its listener on the last accept, so **a pool that fails to reuse gets a REFUSED connect and the test fails in a second instead of hanging**.

Covered: a second request costs no reconnect (asserted on the accept count); `Connection: close`, an HTTP/1.0 response without `keep-alive`, and an unknown-length response each retire the connection; HTTP/1.0 *with* `keep-alive` is reused; a server that closed an idle connection is retried once and succeeds; the cap and the age limit; `dispose` closes every idle connection (two servers, both must read EOF); the pool crosses neither ports nor hosts; the free `fetch` still announces `Connection: close`; and 204 / HEAD framing both end-to-end and at the wire level with a scripted reader that proves the following response is *carried*, not swallowed as a body.

Two measurement notes, both learned the hard way: the accept counters live in a `ref(struct(...))` because a module-level mutable written across a suspension point reads back as zero, and "did `dispose` close every idle connection?" is answered by what the PEERS observe — a closed connection is one the server at the other end reads EOF from, so its task finishes, while one still open leaves that server parked on a read until the deadline reports it. Two descriptor-table oracles were tried and both are dead ends: a lowest-free-descriptor probe reads the same number whether the socket leaked or not (a listener closed earlier in the test leaves a hole below it), and the exact open-count that fixes that (`dup(0)` N times ⇒ `highest + 1 == open_count + N`) is not portable — a Windows socket is not in the CRT descriptor table at all, and the count did not move on the intel macOS runner either. Peer observation needs no platform knowledge, and it is what every other test in the group already depended on implicitly, since each ends in a `task.await(io)` that only returns once the client's socket is really closed.

## The server half is NOT here, and the reason is measured

`std/async`'s `timeout` is a blocking-poll plain `fn` that drives the event loop itself, so calling it inside `serve_once`'s `io.async` body nests the loop (`issues/fixed/sync-await-in-plain-fn-nests-the-event-loop.md`); every other combinator there has the same shape and takes `JoinHandle`s rather than futures, so there is nothing to compose with the framing read. The pattern that does work inside a task is hand-rolled exactly once in this tree (`_fetch_deadline`) and carries costs that belong in a design decision rather than in this PR: aborting a read parked in the I/O backend leaks its 8 KiB buffer and its state machine per timed-out connection, and server keep-alive changes what `serve_once` MEANS in ways that invalidate existing tests (pipelining becomes correct rather than smuggling).

`plans/backlog/ASYNC_DEADLINE_COMBINATOR.md` names the blocked call site, those costs, three options and the recommendation: add a `Future`-returning `with_deadline` to `std/async`, re-express `_fetch_deadline` in terms of it, then land server keep-alive on top as an **opt-in** — opt-in because keep-alive on a strictly serial server lets one client monopolize it.

## Three defects found and filed

- `issues/a-ref-value-passed-to-an-async-future-is-never-released.md` — **open.** A `ref` value passed as a parameter to an `io.async` future is dup'd into the capture record and never dropped, unbounded (50 calls, 50 leaks, still leaked after the frame returns). It is why `HttpClient`'s `Dispose` — and `TcpStream`'s and `TlsStream`'s — cannot fire on a drop, which is why `close_idle()` is documented as the way to release a client and why the `dispose` test calls it by hand.
- `issues/a-module-global-is-lost-across-an-async-suspension.md` — **open.** A module-level mutable read or written across a suspension point in an `io.async` body operates on a copy. It reads back as zero, which looks exactly like "the code under test never ran".
- `issues/fixed/a-dropped-tlsstream-leaks-its-socket-and-openssl-objects.md` — **fixed here.** `TlsStream` had no `Dispose`, so its socket and its OpenSSL `SSL`/`SSL_CTX` could not be released except through the async `close(io)`.

Each has a runnable reproducer under `issues/repros/`.

## Verification

- `yo check ./std --std-path ./std` — 173/173.
- `tests/http/http.test.yo` 51 pass, `server.test.yo` 13, `http_limits.test.yo` 6.
- Hollow-green probe done: flipping one assertion to `assert(false, …)` fails the file. The `dispose` test is verified red-first too — with its `dispose` call removed it fails with "the first server never saw its connection close: timeout: the deadline elapsed".
- `tests/http/wire.test.yo` type-checks but cannot be COMPILED by the installed release seed — its scripted reader is the regression exercise for #547's own `src/codegen/exprs/atom.yo` capture fix, which the seed predates (verified identical on the base commit, in a scratch worktree). CI builds stage-1 first, so it compiles there; the new scripted-reader cases were added to `http.test.yo` instead, written capture-safely so they run locally too.
- `yo fmt --check` clean on every touched file.

## No user docs

There is no HTTP page under `docs/en-US/` or `docs/zh-CN/` (nor any networking page), so there was nothing to extend and I did not invent one. The user-facing documentation is the doc comments, which `yo doc` renders.

---

## Update 2026-09-14 — root-caused, fixed, and cleaned for release

**The blocker was a compiler bug, in neither the HTTP client nor either I/O backend.** A `while` whose body awaits, inside a `match` arm, emitted the statements that FOLLOW it *inside* the loop body, replaying them every iteration. Fixed in #677 (merged); full write-up in `issues/fixed/a-while-loop-inside-a-match-arm-runs-its-trailing-code-every-iteration.md`.

`_fetch_deadline` is exactly that shape — its race loop is in the `.Some(limit)` arm and the statement after it is `cond(h.is_finished() => deliver, true => { h.abort(); throw Timeout })`. Inside the loop, that ran on the **first turn**, found the exchange unfinished because it had barely started, aborted it and threw `HttpError.Timeout` — on an exchange that completes in about a millisecond.

That accounts for every open question this PR's issue accumulated over days of investigation:

- **always `Timeout`, never the real error** — the fallback arm is reached unconditionally on turn one.
- **the server sits in `accept`** — the client aborts before connecting.
- **identical on kqueue and io_uring** — neither is involved.
- **"the parent await never resumes"** — it was aborted, not stuck.
- **platform roulette** — the abort is deterministic; only how far the exchange gets first is not.

It also **reproduces locally**, contrary to this issue's record of 27 clean runs — that was measured against a runtime with two since-fixed wake-path bugs.

### The investigation checkpoints are gone

The branch carried **eleven `eprintln` checkpoints inside `std/http/client.yo` and `std/http/wire.yo`** — `[tr] read entered`, `[wire RSP] issuing read`, `[clt] parent await resumed`. They print on every read of every HTTP request in every program that links `std/http`, and must not ship. Removed, with the `_role` binding that existed only to tag them, their DIAGNOSTIC comment blocks, and the ten matching checkpoints in `tests/http/http.test.yo`. The `eprintln`s that remain there are real failure reporting.

### `_fetch_deadline`'s spin is left as-is, deliberately

It is now *correct* — it was only ever broken by the codegen bug. `plans/backlog/ASYNC_DEADLINE_COMBINATOR.md` argues it should become a real combinator over `Park`/`Waker` rather than a spin, and it should — on its own merits, after the release, not as a rushed change inside a release window.

### Verification

- `tests/http` — **77/77** with the checkpoints stripped.
- Full fast suite on the rebased branch — reported below.
- Rebased onto develop `2a7bc4269`, which carries #677 plus the three async runtime fixes (waker-token UAF, per-thread event-loop leak, forall binder identity) and both async I/O audits.

### DIFF
--- a/issues/a-bodyless-http-response-is-not-read-until-the-deadline.md
+++ b/issues/a-bodyless-http-response-is-not-read-until-the-deadline.md
@@ -215,6 +215,53 @@ Already checked and NOT it:
   non-blocking kernel entry
   (`issues/fixed/io-uring-defer-taskrun-poll-never-enters-kernel.md`).
 
+## After the rebase onto develop (2026-09-12): both Linux legs PASS
+
+The branch was ~30 commits behind — it predated #590 (`JoinHandle.abort`
+cancels the I/O the task is suspended in, which rewrote the future header and
+`__yo_io_cleanup`), #592 (six async-emitter dispatch fixes) and #598 (a
+static-dispatch call's C return type). Rebased onto `4e541f82a` and re-run as
+**34663406193**:
+
+| leg | before | after |
+| --- | --- | --- |
+| `test (ubuntu-latest)` | FAIL | **success** |
+| `test (ubuntu-24.04-arm)` | FAIL (the tagged trace above) | **success** |
+
+So something in those thirty commits moved it. That is NOT proof it is fixed —
+the whole issue is timing-dependent, and two green legs on one run is exactly
+what "timing-dependent" looks like half the time. It IS enough to say the next
+trace must be taken on the rebased branch, and that the analysis above was
+performed against a compiler three async-fix batches old.
+
+**The full-corpus hollow sweep still fails the same two tests**, but with a
+DIFFERENT signature, and that difference matters:
+
+```
+✗ HttpClient: a 204 ends at its headers, whatever its Content-Length says
+    Test failed with exit code 134
+✗ HttpClient: a HEAD response ends at its headers and the connection is reused
+    Test failed with exit code 134
+```
+
+Exit 134 is SIGABRT, which IS this bug's signature — the deadline throws,
+the test's `exn` handler `assert(false, ...)`s, and that aborts. The two
+children printed no checkpoints, but that is a property of the harness and not
+evidence about the run: the runner captures each test's output and REPLAYS it
+after the verdict line (in the same log, a passing test's `[wire RSP]` lines
+appear *below* its `✓`), and a child that aborts loses the capture. So the
+sweep is the same failure, not a second one, and it still reproduces in that
+binary shape on a leg whose sibling `test (ubuntu-*)` legs passed — which is
+the clearest statement yet of how narrow the timing window is.
+
+Worth ruling out anyway while reading the file: the keep-alive tests bind
+19840–19855, one port each, so the two failures are not colliding with each
+other or with their predecessor.
+
+Making the sweep keep an aborted child's output would pay for itself here; it
+runs each file through `BIN test <file> --parallel 1` with no `--std-path` and
+no `--verbose`.
+
 ## Earlier hypothesis, now refuted: a two-arm match with an await in each arm
 
 `std/http/client.yo`'s `_Transport.read` is an `io.async` body whose only
--- a/issues/a-bodyless-http-response-is-not-read-until-the-deadline.md
+++ b/issues/a-bodyless-http-response-is-not-read-until-the-deadline.md
@@ -57,7 +57,44 @@ Ruled out locally so the next session does not repeat them:
 - An x86_64 build under Rosetta could not be produced — homebrew's OpenSSL is
   arm-only, so `--target x86_64-apple-darwin` fails to link `TLS_client_method`.
 
-## The most likely cause: the client's read is a two-arm match with an await in each arm
+## Three suspects EXCLUDED by measurement (2026-09-11)
+
+Recorded so none of them is re-investigated. Each was excluded by comparing
+emitted C or by reading the registration flags, not by a passing test.
+
+1. **The two-arm `_Transport.read` dispatch — NO.** The shape (an enum with two
+   payload arms, each `n = e.io.await(stream.read(...), e)` over
+   differently-shaped futures) was written standalone and emitted with a
+   pre-#592 compiler and with #592's: the C is BYTE-IDENTICAL, and both
+   binaries compute the right answer over 200 alternating iterations. None of
+   #592's six dispatch fixes touch a plain two-arm point.
+
+2. **An edge-triggered readiness registration that misses an edge that already
+   happened — NO, not on macOS.** `__yo_io_register_kevent`
+   (`src/codegen/async/runtime_io_macos.yo`) issues
+   `EV_SET(..., EV_ADD | EV_ONESHOT, ...)` with **no `EV_CLEAR`**, so
+   `EVFILT_READ` is LEVEL-triggered: registering it against a socket whose
+   buffer already holds the response fires immediately.
+
+3. **The while-loop family — NO.** `read_http_message_buffered`'s read loop is
+   an outer `while` whose await is its first statement, with a NESTED `while`
+   (no await) inside a `cond` arm after it, both testing the same `done` flag.
+   That combination was compared function-by-function between a pre-#592
+   compiler and #592+#593: **362 functions on both sides, 0 bodies differ**.
+   The only raw-text difference in the whole file is one unreferenced struct
+   slot (`__yo_t1 var_N; // io`, 0 uses) that the post-fix suspension analysis
+   stops declaring. No store, no read, no branch.
+
+   Worth noting for any future A/B: the compiler's own emission never includes
+   this function, because `std/http` is not in `src/main.yo`'s import closure.
+   An emit A/B over the compiler tree therefore says nothing about it, and the
+   comparison has to be driven from a small program that imports the module.
+
+**That leaves the platform I/O completion path.** The client-side checkpoints
+now on this branch answer the remaining question directly: whether the read is
+never ISSUED, or issued and never woken.
+
+## Earlier hypothesis, now refuted: a two-arm match with an await in each arm
 
 `std/http/client.yo`'s `_Transport.read` is an `io.async` body whose only
 await sits inside a `match`, once per arm:
@@ -88,9 +125,10 @@ It also explains the timing-dependence. Which arm's continuation is emitted as
 the representative, and whether the loop delivers the completion inline or
 through the ready queue, decides whether the mis-routed path is taken.
 
-**So the first thing to do after #592 lands is re-run this branch's battery.**
-If the hang is gone, this issue closes with #592 rather than needing a runtime
-fix. If it survives, the runtime question below is the next one.
+**This was wrong** — see exclusion 1 above. #592 landed and the shape it
+would have had to fix emits identically on both compilers. Kept because the
+reasoning is the right reasoning for the NEXT dispatch-shaped suspicion; only
+the conclusion was wrong.
 
 ## Where to look, if #592 does not fix it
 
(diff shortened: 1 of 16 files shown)


shd101wyy:
Rebased onto `develop` (now carrying #550, #560, #554, #558, #555, #559 and the two compiler PRs #562/#561). Conflicts resolved in `std/http/client.yo` (the doc sweep's module header vs this PR's; merged, with the `## Stability` section rewritten for the pool), `plans/README.md` and `plans/STD_API_STABILIZATION.md` (both lists kept).

## On the red legs

`test (ubuntu-latest)` and `test (macos-26-intel)` timed out on **"HttpClient: a 204 ends at its headers, whatever its Content-Length says"**, and the hollow sweep flagged `tests/http/http.test.yo RED` for the same reason. Passing on some legs and hanging on others is exactly the shape a framing bug takes — whether the peer's writes coalesce is the peer's decision — so the first thing to settle was whether the framing is right.

**It is, and now there is a test that says so without a socket.** `tests/http/wire.test.yo` already had a `_ScriptedReader` that hands out at most `chunk` bytes per call and counts its calls; two new tests drive the body-less path through it:

- a 204, a 304 and a HEAD response are each framed at their header section at **chunk sizes 1** (every header split), **42** (a split inside the CRLFCRLF that ends the header section) and **4096** (all at once) — with the same HEAD bytes read as a plain `Response` as the contrast, proving the decision is the status/kind and not the headers;
- a 204 followed immediately by another response **in one read**: the 204 is exactly its header section, the next response is carried whole, and reading it costs no read.

All seven wire tests pass, including at one byte per read.

**And the socket exchange is not slow.** Locally the same 204 exchange completes with the deadline set to **20 ms**. A 10-second deadline expiring on a runner is therefore a hang, not a loaded box — which is worth stating plainly rather than papering over with a bigger number, so the test keeps its original 10 s.

Local verification with the compiler built from the new `develop`:

| | |
| --- | --- |
| `yo check ./std --std-path ./std` | 175/175 |
| `tests/http/http.test.yo` | 51/51 (run three times, and twice back-to-back) |
| `tests/http/server.test.yo` | 13/13 |
| `tests/http/wire.test.yo` | **7/7** (+2) |

Re-running CI on the rebase; if the same two legs go red again that is a second data point on a real hang, and the wire tests have already ruled the framing out of it.

shd101wyy:
Second data point: the same test timed out again, this time on **both Linux legs** (`ubuntu-latest` and `ubuntu-24.04-arm`). It passes on macOS/arm64 locally, run repeatedly and twice back to back.

Two candidates are now ruled out rather than assumed:

- **Framing.** The new `tests/http/wire.test.yo` cases frame a 204, a 304 and a HEAD response at their header section at chunk sizes **1** (every header split), **42** (a split inside the CRLFCRLF that ends the header section) and **4096**, and carry the message that follows a 204 in the same read. 7/7.
- **Slowness.** The same socket exchange completes locally with the deadline set to **20 ms**. A ten-second deadline expiring is therefore a hang, not a loaded runner — so raising the number would only change how long CI waits to tell us the same thing.

What is left is *where* it stops, and with no Linux box on this machine only CI can say. So the test now prints a checkpoint either side of each await — server spawned, response one (status, body length, and the server's accept/request counters), the idle count that decides pooling, request two on the pooled connection, response two, server task joined. `eprintln` rather than `println`, because a buffered `println` is lost when the deadline throws, and the runner passes `--verbose`.

The next run's log should say whether the first exchange completes at all, whether the server ever accepted, and whether the hang is on the pooled second request — which is the difference between a client-side framing decision, a server-side one, and something in the epoll backend that the kqueue backend does not do.

This PR stays open until that is answered; everything else in it is verified locally (`yo check ./std` 175/175, `tests/http/http.test.yo` 51/51, `server.test.yo` 13/13, `wire.test.yo` 7/7, `fmt --check` clean).

PR #558, opened 2026-09-10
std/encoding: the module-prefix stutter, with deprecated aliases for one release

shd101wyy:
Closes the module-prefix-stutter row of D2 (naming) in
`plans/STD_API_STABILIZATION.md` — 6 of the 7 sites, with the 7th deferred for
a stated reason.

## The rule

A function in module `std/encoding/json` is reached as `json.parse(...)` once
the module is imported as a module. That is how this tree already reads
`std/sys/events` (`events.fs_event_stop(...)`) and how Rust reads
`serde_json::from_str`. The `json_` in `json.json_parse` said the same word
twice at every call site.

## What was renamed — 33 functions, eight modules

| module | old → new |
| --- | --- |
| `encoding/json` | `json_parse`/`_bytes`/`_string`/`_exn`/`_bytes_exn`/`_string_exn` → `parse`/`parse_bytes`/`parse_string`/`parse_exn`/`parse_bytes_exn`/`parse_string_exn`; `json_stringify`/`_pretty` → `stringify`/`stringify_pretty`; `json_encode`/`json_decode` → `encode`/`decode` |
| `encoding/csv` | `csv_parse`/`_with`/`_strict` → `parse`/`parse_with`/`parse_strict`; `csv_write`/`_with` → `write`/`write_with` |
| `encoding/base64` | `base64_encode`/`_url` → `encode`/`encode_url`; `base64_decode`/`_url`/`_exn`/`_url_exn` → `decode`/`decode_url`/`decode_exn`/`decode_url_exn` |
| `encoding/hex` | `hex_encode` → `encode`; `hex_decode`/`_exn` → `decode`/`decode_exn` |
| `encoding/html` | `html_encode`/`html_decode` → `encode`/`decode` |
| `encoding/toml` | `toml_parse` → `parse` |
| `encoding/percent` | `percent_encode` → `encode`; `percent_decode`/`_bytes` → `decode`/`decode_bytes` |
| `encoding/utf16` | `utf8_to_utf16` → `from_utf8`; `utf16_to_utf8`/`_exn` → `to_utf8`/`to_utf8_exn` |

## How it lands without breaking callers

The short name is the real definition; the old prefixed name stays as a **thin
deprecated alias**, marked in its doc comment with the release that removes it
(v0.2.32). That is the mechanism `json_parse_result` and `derive(ToString)`
already used after the v0.2.28 breaking window.

Every in-tree call site moved: `std/`, `src/`, `tests/`, `docs/en-US` and
`docs/zh-CN`, and the `test("...")` **name strings** — a rename that leaves the
test name saying the old spelling is a filed defect in this tree
(`issues/collection-test-names-still-use-pre-rename-method-spellings.md`),
because `--test-name-pattern` is the documented way to run one test.

## `glob_match` is deferred, with the reason written down

`glob.match` is **unspellable** — `match` is a Yo keyword — so the rename would
have to be to a DIFFERENT name (`glob.matches`, Rust's `Pattern::matches`), and
`GlobPattern.matches` already exists one screen below it. Two `matches` in one
module is a decision, not a cleanup. Worse, the one std caller
(`std/fs/walker.yo`) itself exports a function called `glob`, so
`glob :: import("../glob")` collides head-on there.

`plans/backlog/MODULE_PREFIX_STUTTER_REMAINDER.md` records that, the measured
remainder of every other module whose exports repeat its name, and the
counter-rule: `std/assert`'s `assert_eq` and `std/sys/clock`'s `clock_gettime`
are NOT stutters — nobody imports those as modules, and the second is a C
binding whose value is being spelled exactly as `clock_gettime(2)` is.

## Coverage

- `yo fmt --check ./std ./src ./tests` clean
- CI's full suite across six platforms is the gate for a rename sweep, per
  `.github/instructions/testing.instructions.md`; the alias layer means the
  test count should be unchanged.


shd101wyy:
Rebased onto `develop`, which now carries #554's real TOML parser. The rebase conflicted in `std/encoding/toml.yo` because this PR renamed `toml_parse`/`toml_stringify` on the OLD line-based implementation, which no longer exists — so the rename was re-applied by hand to the new byte scanner:

- `parse` and `stringify` are the real definitions; `toml_parse`/`toml_stringify` stay as thin deprecated aliases marked for removal in v0.2.32, matching what the other seven modules in this PR do.
- The module doc's example, `TomlValue`'s `ToString` impl, and every call site in `tests/encoding/toml.test.yo` (which #554 moved out of `tests/toml/`) now use the module-qualified spelling.

Verified locally with the compiler built from the new `develop`:

| | |
| --- | --- |
| `yo check ./std --std-path ./std` | 174/174 |
| `tests/encoding/toml.test.yo` | 36/36 |
| `tests/std_export_coverage.test.yo` | 13/13 |
| `tests/encoding/{json,base64,csv,hex,html,percent,utf16}.test.yo` | 68 / 21 / 7 / 10 / 15 / 8 / 15 |
| `tests/crypto/sha256.test.yo` | 12/12 |

PR #559, opened 2026-09-10
std: document 359 undocumented members, and correct what the docs row said

shd101wyy:
Closes most of the `docs` row in `plans/STD_API_STABILIZATION.md` §1, and
corrects that row's diagnosis.

## The row was wrong about the cause

§1 said "`//` instead of `///` (dropped by `yo doc`): atomic.yo ~130, log.yo
15, …". Measured against the code, almost none of it was that. In the worst
offender — `std/sync/atomic.yo`, 33 `//` lines — every one of those lines is a
SECTION BANNER, and its 126 members simply had **no comment at all**. A
mechanical `//` → `///` conversion would have converted the wrong lines and
documented nothing.

The metric that means something is impl members with no preceding `///`:

| | members with no doc |
| --- | --- |
| before | **838** |
| after | **479** |

`atomic.yo`'s 126 land in their own PR (beside the `MemoryOrder.Consume`
removal); the 359 here cover `std/collections/*`, `std/imm/*`, `std/crypto/*`,
`std/encoding/*`, `std/url`, `std/time/*`, `std/fs/*`, `std/sys/*` and the root
modules, plus `## Stability` sections on the modules swept.

## What a doc comment says here

House style, matching `std/collections/ordered_map.yo`, `std/io/index.yo` and
`std/time/instant.yo`: what the thing answers in terms a caller thinks in, the
Rust counterpart where there is one, and the facts a caller can otherwise only
get by reading the body — what PANICS and when, what is O(n), what invalidates
a pointer or an iterator, what a `.None`/`.Err` arm actually means, which of
two similar methods to reach for. A deliberate divergence from Rust says so and
why, which is the most valuable sentence in any of these files.

## Scope

**Docs only.** No behaviour, signature or name changes, so this stays
reviewable as a docs change.

## Still open, now measured rather than estimated

479 members, and **16** modules with no `## Stability` section: `allocator`,
`path`, `signal`, `error`, `thread`, `term`, `prelude`, `build`, `env`,
`assert`, `sys/file`, `sys/dir`, `sys/dns`, `sys/errors`, `sys/externs`, and
`sync/atomic` (marker in its own PR). §1's "8 modules" was a subset — the ones
a reader is most likely to reach for, not the ones that lack a marker.

## Coverage

- `yo check ./std --std-path ./std` → 173/173 (a stray `///` where no
  definition follows is a parse error, so this is the real gate)
- `yo fmt --check ./std` clean


shd101wyy:
Rebased onto `develop` (now carrying #560's imm/Vec doc correction, #554's real TOML parser, #558's module-prefix rename and #555's `Stream`), and the CI failure this PR had is fixed.

**What was red:** `Self-hosted test subcommand (yo-self tier-1 gates)` → GATE 7, `lsp-member-definition` GOLDEN-DIFF. That case pins absolute LINE NUMBERS inside `std/collections/array_list.yo` — `push` at 244, `new` at 29, `ArrayList` at 17 — so adding `///` comments to that module moves all three. The server's answers are still correct; they point at 269, 39 and 26 now. Re-recorded in its own commit.

**Conflicts resolved,** all of them "develop is newer, keep develop, keep this PR's `## Stability` section":

- `std/encoding/json.yo`, `std/encoding/toml.yo` — the sweep's examples used the pre-rename `json_parse` / `toml_parse` spellings and described the line-based TOML subset parser that #554 deleted. develop's text kept.
- `std/imm/vec.yo` — the sweep called it "Persistent immutable vector", which is the claim #560 corrected. Merged: develop's "flat copy-on-write array, NOT a structurally-shared trie" framing, plus this PR's `own(self)`-moves-the-receiver paragraph, which develop did not have.
- `std/io/index.yo` — develop's "STILL TO COME" is newer (the async iterator protocol landed in #555, so the remaining blocker is the throwing read); this PR's `## Stability` section kept and its stale `lines()` sentence pointed at the real blocker.
- `std/sync/channel.yo` — the sweep restated the spawn-capture caveat that #562 made untrue. Dropped; the `## Stability` section kept.
- `plans/STD_API_STABILIZATION.md` — the doc-sweep record kept ahead of §5's new heading.

Verified locally with the compiler built from the new develop:

| gate | result |
| --- | --- |
| `scripts/bootstrap/gates_fast.sh` GATE 0–8 | **`=== T1_DONE (ds2) failures=0 ===`** |
| `yo check ./std --std-path ./std` | 175/175 |
| `tests/std_export_coverage.test.yo` | 13/13 |
| `tests/sync/channel.test.yo` | 41/41 |

PR #560, opened 2026-09-10
std: the three §5 maintainer decisions — imm/Vec doc, MemoryOrder.Consume, HashMap keys

shd101wyy:
`plans/STD_API_STABILIZATION.md` §5 held three shape questions open "for the
maintainer, not an agent". All three are answered here, each with the
measurement that decided it. Two changed code; one deliberately did not.

## 1. `imm/Vec`: flat COW, documented as such — no RRB

The doc claimed **"structural sharing"**, which was false about the
implementation. The backing store is one flat array behind an atomic refcount:

| operation | refcount 1 | refcount > 1 |
| --- | --- | --- |
| `get` / `len` | O(1) | O(1) |
| `push` / `pop` / `set` | O(1) amortized, in place | **O(n)** — the array is copied |

The module header now carries that table and the reason the trade was kept. The
ergonomics already push callers onto the fast path — `push` takes `own(self)`,
so the receiver is MOVED and keeping the earlier version needs an extra binding
(`kept := v; bigger := v.push(x)`), which is exactly what raises the refcount
above one and sends `push` down the copying path. And a flat array is what
makes `Vec` right for the case it is actually reached for: build once, then
read and share, with O(1) indexing as a single pointer offset. `imm.List` is
the answer for cheap prepend on a shared value; `ArrayList` for mutation. RRB
stays possible behind this same API, unchanged, if a workload ever needs shared
mutation to be sublinear.

## 2. `MemoryOrder.Consume`: removed — and it surfaced a real bug

C11 has `memory_order_consume` and **no production compiler implements it**:
clang and gcc both strengthen it to `acquire`, and C++ formally discouraged it
(P0371R1) for that reason. A variant that costs `acquire` while promising
dependency ordering is a correctness trap — code written against the promise is
unsound on any compiler that ever keeps it, and nothing today does. Rust omits
it.

Measured before removing: **zero uses in `std/` or `src/`**. The only eleven
were in `tests/sync/atomic.test.yo`, as an arbitrary order for `fetch_xor`.

**The bug it surfaced:** `Atomic*.load` accepted a STORE-only order and `store`
accepted a LOAD-only one, passing it straight through to C11. §7.17.7.1–2
forbid both, but the order arrives as a **runtime value** here, so the C
compiler cannot reject it the way it rejects a literal —
`a.load(MemoryOrder.Release)` compiled clean at `-O2`, ran, and silently got
whichever ordering the runtime switch fell into.

Rust PANICS on exactly these combinations rather than clamping, because
substituting a different ordering hides the bug. So does this now, at all 33
sites (11 `load`, 11 `store`, 11 compare-exchange failure orders).
`_read_order`'s existing clamp for a compare-exchange loop's failure order is
untouched: there the mapping is *required* by C11, not a guess.

`std/sync/atomic.yo` also gained the `## Stability` section it lacked and `///`
docs on **all 126** of its members — every one had no comment at all, which is
why §1's docs row (which blamed `//` instead of `///`) was misdiagnosing this
file: its 33 `//` lines are all section banners.

## 3. `HashMap.with_random_keys()`: not shipped; the recipe is documented

`with_keys(k0, k1)` already does the whole job. The wrapper was **probed, not
assumed**, and the probe decided it: adding
`{ random_u64 } :: import("../crypto/random")` to
`std/collections/hash_map.yo` takes `yo check ./std` from **173/173 to
166/173** — `std/crypto/random` imports `collections/array_list`, `process`,
`string`, `encoding/hex`, `error` and `fmt`, and a core collection cannot
depend on that closure.

The signature is wrong for a constructor too: `random_u64` takes an
`Exception`, so the convenience would be `with_random_keys(exn)` and a
`HashMap.new()`-shaped call would throw on entropy failure. Rust's
`RandomState` can abort, but Rust's randomness is not in the collection's
dependency graph.

So the recipe stays at the call site, where a reader needs to SEE that the
choice was made:
`HashMap(K, V).with_keys(random_u64(exn), random_u64(exn))`. `new()` stays
deterministically keyed — the bootstrap fixpoint gate depends on byte-identical
emitted C, so it could not be randomized even if that were wanted.

## Coverage

- `cli-diff-test.sh` scorecard **PASS 96 GOLDEN-DIFF 0 NO-GOLDEN 0**, including
  a new `atomic-illegal-load-order-panics` case: the program compiles cleanly,
  `build run` propagates rc 1, and the golden pins the diagnostic. Verified by
  hand first — `a.load(MemoryOrder.Release)` aborts with the message at
  `std/sync/atomic.yo:166`.
- `tests/sync/atomic.test.yo`'s eleven `Consume` uses moved to `AcqRel`, which
  is legal for every operation they appear in.


shd101wyy:
Rebased onto `develop` and verified locally: `yo check ./std --std-path ./std` 174/174, `tests/sync/atomic.test.yo` 71/71, `tests/collections/hash_map.test.yo` 84/84. Also added the doc half this PR was missing — `docs/en-US/THREAD_SAFETY.md` and `docs/zh-CN/THREAD_SAFETY.md` still listed `MemoryOrder.Consume` in the enum and said nothing about the illegal-order panic.

PR #561, opened 2026-09-10
std/async: a real waker/park primitive, and `yield_now` without the 1 ms floor

shd101wyy:
An async task waiting for something the event loop does not know about had
exactly one tool: `yield`, implemented as a 1 ms timer. Every hand-off cost a
millisecond of wall clock even when the peer was ready immediately, and a task
waiting on another task's progress could only poll. This adds the primitive the
runtime was missing.

## Runtime — `src/codegen/async/runtime_core.yo`

| symbol | what it does |
| --- | --- |
| `__yo_async_park_start` | a pending future no I/O source will ever complete — only a wake can advance it |
| `__yo_waker_new` | takes a reference on that future, bumps a live-waker count |
| `__yo_waker_wake` | CAS state 0 → -1, then enqueue the continuation |
| `__yo_waker_is_woken` | reads the state |
| `__yo_waker_release` | drops the reference |
| `__yo_async_yield_start` | a pending future pushed onto `__yo_yield_pending` |
| `__yo_async_drain_yields` | completes that whole list |

A wake that lands **before** the awaiting task registers is not lost: the
await-point protocol reads `state` before it registers a continuation, so the
early wake is observed at registration time.

`__yo_async_drain_yields` runs inside `__yo_async_run_ready_tasks` **after** the
budget snapshot, so a task that yields can neither starve I/O nor re-enter its
own turn.

All three loop drivers (`poll_step`, `run_until_complete`, `wait_all`) now
consult `__yo_async_has_pending_yield()` / `__yo_async_has_pending_work()`, so a
loop whose only remaining work is a yield or a live waker does not report itself
idle — and a loop with a parked task and **no** waker left alive reports a wake
deadlock instead of hanging.

The state and the forward declarations live in the flags block because the loop
drivers, emitted earlier in the file, call them.

## Std — `std/async/waker.yo`

- `Waker` — `wake()`, `is_woken()`, and a `Dispose` that releases the reference.
- `Park` — `new()`, `waker()`, `wait(io)`: the awaiting half.
- `park(register, io)` — the combinator: hand the waker to `register`, then await.
- `yield_now(io)` — one loop turn, no timer.

`std/async/index.yo`'s `yield` is deliberately **left on its 1 ms timer**:
pointing it at the new extern fails to LINK under the v0.2.30 seed, which emits
a runtime without the symbol. The switch is a two-step across releases and is
recorded as seed-gated in the plan.

## Tests — `tests/async/waker.test.yo` (8)

Early wake is not lost; wake is idempotent; a spawned task parks and is woken;
one signal wakes eight waiters; a dropped waker does not hang the loop; a
400-hand-off relay completes; poll-until-finished over `yield_now` still polls
I/O; `yield_now` runs every turn.

A 400-hand-off relay measures **0 ms** over `yield_now` against **603 ms** over
`yield`. Inside a test batch the same measurement is misleading for a reason
that has nothing to do with the waker, filed separately as
`issues/yield-now-costs-a-millisecond-per-turn-inside-a-test-batch.md` so it
cannot mask future async perf work.

## Verification

| gate | result |
| --- | --- |
| `yo test ./tests --exclude tests/internal --exclude tests/cli-cases` | 4000 passed / 0 failed |
| `scripts/bootstrap/gates_fast.sh` GATE 0–8 | `T1_DONE failures=0` |
| `scripts/bootstrap/fixpoint_only.sh` | `FIXPOINT_HOLDS`, stage2 hollow=0 |
| `tests/async/waker.test.yo` | 8/8 |

Docs updated in both languages (`docs/en-US/ASYNC_AWAIT.md`,
`docs/zh-CN/ASYNC_AWAIT.md`); plan record appended to
`plans/STD_API_STABILIZATION.md`.

### DIFF
--- a/docs/en-US/ASYNC_AWAIT.md
+++ b/docs/en-US/ASYNC_AWAIT.md
@@ -885,6 +885,69 @@ main :: (fn(io : Io) -> unit)({
 export main;
 ```
 
+### Waking a task from another task: `Waker` and `Park`
+
+`yield` hands the loop a turn; it does not let one task wait for *another
+task's progress*. That needs a token the other side can fire, and
+`std/async/waker` is it.
+
+```rust
+{ Park } :: import "std/async/waker";
+
+// The waiter. Create the park, hand its waker to whoever will signal, then
+// suspend — in that order, and with no await in between.
+p := Park.new();
+waiters.push(p.waker());
+io.await(p.wait(io), io);
+
+// The signaller, from any other task:
+match(waiters.pop(), .Some(w) => w.wake(), .None => ());
+```
+
+Three properties are worth knowing, because primitives built on this depend on
+them:
+
+- **A wake that arrives before the sleeper suspends is not lost.** An await
+  point reads the future's state before it registers a continuation, so an
+  already-woken park resumes inline instead of suspending.
+- **Waking is idempotent.** A second `wake()`, or a wake of a park whose task
+  has already finished, does nothing. A waiter list can therefore signal
+  everyone without tracking who already ran.
+- **A park that nothing can wake is reported, not hung.** The runtime counts
+  live waker tokens; when nothing is runnable, no I/O is outstanding, and a
+  token is still alive, the loop says so and stops instead of spinning.
+
+`park(register, io)` wraps the whole sequence for the common case of a single
+waker:
+
+```rust
+{ park } :: import "std/async/waker";
+
+io.await(park((w : Waker) => { slot.* = Option(Waker).Some(w); }, io), io);
+```
+
+`register` runs *before* the suspension and receives the waker, so the order
+cannot be got wrong. This is the shape Rust's `Future::poll(cx)` uses. Reach
+for `Park` directly when a waiter list wants to hold the token itself.
+
+### `yield_now`: a fairness yield with no timer
+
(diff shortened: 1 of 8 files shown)


shd101wyy:
Rebased onto `develop` (which now carries #562, the five compiler fixes — including the spawn-capture release and the dup/drop cancellation guard, both of which touch RC emission around closures) and re-verified locally from scratch:

| gate | result |
| --- | --- |
| `yo test ./tests --exclude tests/internal --exclude tests/cli-cases` | **4064 passed / 0 failed** |
| `scripts/bootstrap/gates_fast.sh` GATE 0–8 | **`=== T1_DONE (wk2) failures=0 ===`** |
| `scripts/bootstrap/fixpoint_only.sh` | **`FIXPOINT_HOLDS`** — stage2 hollow=0, STAGE2_RC=0, CLANG_RC=0, STAGE3_RC=0 |
| `tests/async/waker.test.yo` | 8/8 |
| `tests/async/channel.test.yo` | 15/15 |
| `tests/arc.test.yo` | 15/15 |
| `tests/thread.test.yo` | 13/13 |

The re-run matters because #562 changed how a closure's captures are counted, and `Waker`/`Park` are exactly closure-captured RC handles.

PR #562, opened 2026-09-10
compiler: five defects — escape scanning, a spawn-capture leak (and the dup/drop cancellation that made it unsafe), a lost module global, a silently-zero array length

shd101wyy:
Five bugs. Four were found while writing or measuring std, each silent, each
with its own test; the fifth was found by fix 2's own gate and is what makes
fix 2 safe. Four are fixes; one turns a silently-wrong type into a diagnostic
that names the missing feature.

## 1. A unicode escape accepted non-hex digits, and templates ignored `\u`

`\uXXXX` read the next four runes unconditionally and folded a NON-hex rune as
if it were `0`. `"\uZZZZ"` decoded to a single U+0000 byte; `"\u{41}"` — whose
four runes are `{`, `4`, `1`, `}` — decoded to U+0410 instead of U+0041. Both
silent, from a decoder whose own doc comment claimed "JSON.parse semantics",
which THROWS on a non-hex digit. And the template-string escape table had no
`\u` arm at all, so a backtick template kept the escape as literal source text
while the double-quoted form decoded it: one spelling, two meanings.

- `src/utils.yo` — the shared scanner: `hex_digit_value` (with a real failure
  channel), `UnicodeEscape`, `scan_unicode_escape`. Accepts BOTH `\uXXXX`
  (JSON's four digits, with surrogate-pair combining) and `\u{X..XXXXXX}`
  (Rust's braced form, rejecting >U+10FFFF and lone surrogates).
- `src/lexer.yo` — VALIDATES on the double-quoted path (keeping the text raw so
  `yo fmt` still round-trips it, and so the diagnostic carries a position) and
  DECODES on the template path.
- `src/evaluator/values/string.yo` — drops its local `_hex_digit_val`/`_hex4`
  for the shared scanner, so the two literal forms cannot disagree.

A malformed escape is a LEXICAL error, as in Rust, so the gate is eight tests
in `tests/internal/lexer.test.yo`. A cli-case cannot host it: CI's
`fmt --check` scans the fixture trees, and a fixture that does not lex fails
that gate. `tests/string/string.test.yo` pins the well-formed forms — both
spellings, both literal forms, control bytes, multi-byte characters, an astral
code point and a surrogate pair.

## 2. A spawn closure's captures were never released

The wrapper emitted its capture drop only when `get_drop_function_for_type`
resolved a `___drop` C function, and the self-hosted compiler synthesizes none
for an ANONYMOUS closure-capture struct. Both the call-site dup and the wrapper
drop took their `.None` arms, so the wrapper degenerated to "call the closure,
free the memory" and the `+1` taken when the capture struct is built had no
matching decrement: one leaked reference per RC'd capture field, every spawn,
unbounded in a loop (~344 B per `spawn` with a captured `Channel(bool)`).

`_emit_capture_drop_lines` walks the struct's runtime fields with
`generate_drop_code_for_value` — which already handles nested arrays, tuples
and enums — snapshotting `em.code`, unhooking the declared/scope trackers, and
re-emitting the produced lines into the DECLARATION buffer where a spawn
wrapper lives. `src/codegen/functions/declarations.yo` gains the two forward
declarations that made that legal C: `__yo_decr_rc_atomic` and
`__yo_incr_rc_atomic` are defined in the code buffer, so a declaration-buffer
call was an implicit declaration followed by a `static` definition.

No dup is added at the call site: the heap copy is a shallow copy that inherits
the call-site struct's references, so the spawn is a move and exactly one
release balances it.

**This closes `plans/archive/SPAWN_CAPTURE_AUTOCLOSE.md`.** A `Sender` moved
into a `Thread.spawn` closure now closes its channel:
`issues/repros/thread-spawn-capture-blocks-channel-autoclose.yo` reports `true`
on both lines where the second was `false`. The caveats in
`std/sync/channel.yo`, `std/async/channel.yo` and `std/thread.yo`'s
`Pool.join_all` are corrected rather than restated.

Gated by a module-level Dispose counter, not a leak checker (CI runs
`detect_leaks=0` everywhere). **Red-first: 0 disposes under the v0.2.30 seed,
1 with the fix.** `tests/thread.test.yo` asserts exactly once, and sixteen
times across sixteen iterations.

## 2b. …and a closure's capture dups were cancellable against the capturing scope's drop

Releasing the captures is only half the accounting, and running the suite with
just that half proved it: `tests/arc.test.yo`'s "Test Arc shared across thread"
and "Test Arc of struct shared across thread" started failing, reading `0`
instead of `42` after `join()`.

`_search_dup_calls` (`src/evaluator/exprs/begin.yo`) already skipped `io.async`
captures, with the right reason on it — *"the SM path needs BOTH the dup and the
scope-exit drop; cancelling them would leak"* — but that guard was written as a
special case for one construct, so every other closure fell through.

A closure definition's `deferred_dup_expressions` are its capture-struct FIELD
INITIALIZERS (`evaluator/values/anonymous_function.yo`, via
`generate_captured_variable_dup_expressions`), not stores into a container that
shares the capturing scope's lifetime. The capture struct is moved into the
closure value and released by whatever owns it — a spawn wrapper, an async
future, a stored `FuncVal` — so its release can happen BEFORE the capturing
scope ends and the local must keep its own count.

Cancelling the pair left the capture struct holding a borrowed alias: a leak
while nothing released it, a use-after-free the moment something did. The
three-thread sibling test passed throughout, for the wrong reason — two capture
dups is more than one, and the optimizer only ever cancelled a single dup.

With the guard the accounting closes: rc 1 on creation, 2 after the capture
dup, 1 after the wrapper's release, 0 at the local's own (no longer cancelled)
scope-end drop. `tests/thread.test.yo` gains "Thread spawn's captured local is
still owned by the caller", pinning both ends against the Dispose counter.

## 3. A module-level global was lost across an async suspension

`_capture_env_variable` hoisted every referenced variable into the
state-machine struct, module globals included. A global became `sm->var_<id>`:
the field starts at zero (the struct is calloc'd and nothing initializes it
FROM the global), every write went to the field, and the state entry then
declared a local with the global's mangled C name reading `sm->var_<id>` —
shadowing the real global. A body that suspended read and wrote a COPY: the
writes were lost, the next call restarted from zero, and the same body with no
`io.await` updated the real global.

Silent, and worse than silent: `.github/instructions/testing.instructions.md`
recommends a module-level counter as the in-language oracle for a leak, and a
counter that reads zero looks exactly like "the thing under test never
happened". It is why a working keep-alive server reported no accepts.

`src/evaluator/shared/suspension_analysis.yo` returns early for a module-level
global, mirroring the guard the CLOSURE-capture path already had in
`src/evaluator/context.yo` — and using the same REGISTRY rather than
`Variable.is_module_level`, for the reason recorded there.

`tests/async_await.test.yo` covers the control (no suspension), the regression,
a second call continuing from the global, and a spawned task.

## 4. An unsubstitutable array length silently became zero

`t_array_var` can only record an IDENTIFIER, so a length that is neither
compile-time known nor a bare variable name fell through to `Array(elem, 0)`.
That is how `-> Array(u8, T.BYTES)` in a blanket impl emitted a signature
returning `Array_uint8_t_0` while the specialized bodies emitted
`Array_uint8_t_1` and `Array_uint8_t_4`. C caught it only because those are
different types; a length-0 result the body also produced would have compiled,
so this was a silently-wrong-type class rather than a reliable error.

`src/evaluator/types/array.yo` rejects it with a diagnostic naming the
limitation and pointing at
`plans/backlog/VALUE_SUBSTITUTION_IN_TYPE_POSITIONS.md` step 2. This is item 9
of the handover's ranked list ("error, not 0"). The feature itself stays open
(`issues/associated-constant-in-a-type-position-resolves-to-zero.md`), and with
it the collapse of `std/prelude.yo`'s ten byte-conversion blocks.

`tests/array.test.yo` carries the rejection plus an over-rejection canary for
every length form that MUST keep working: a literal, a named comptime constant,
`_` inference, and a `generic(N : usize)` parameter at two widths.

## One new defect filed, not fixed here

Writing fix 2's Dispose-counter gate surfaced a **pre-existing use-after-free**:
a named local stored into a struct FIELD loses its own reference, so once the
container dies the local is dangling —
`issues/a-local-stored-in-a-struct-field-is-dangling-after-the-container-dies.md`,
with a no-thread, no-async reproducer. It reproduces identically under the
v0.2.30 seed, and on the plain-`ref` path the GC's deferral masks it into a
correct-looking answer. The fix belongs in `_optimize_dup_drop_pairs` and needs
the dup/drop emit-diff gate, so it is a stacked PR rather than a rider here;
this PR's own test was rewritten to use a module-level counter so it measures
the spawn wrapper rather than that bug.

## Verification

| gate | result |
| --- | --- |
| `yo fmt --check ./src ./std ./tests` | clean |
| `tests/internal/lexer.test.yo` | 56/56 (+8) |
| `tests/string/string.test.yo` | 292/292 (+3) |
| `tests/array.test.yo` | 20/20 (+2) |
| `tests/async_await.test.yo` | 190/190 (+1) |
| `tests/thread.test.yo` | 13/13 (+3) |
| `tests/arc.test.yo` | 15/15 (red on fix 2 alone, green with 2b) |
| all five reproducers | correct output, red-first under the seed |
| full local suite | see the comment below |

Full suite, `gates_fast.sh` and `fixpoint_only.sh` are running locally and will
be reported in a comment.

### DIFF
--- a/issues/a-local-stored-in-a-struct-field-is-dangling-after-the-container-dies.md
+++ b/issues/a-local-stored-in-a-struct-field-is-dangling-after-the-container-dies.md
@@ -1,0 +1,111 @@
+# A local stored into a struct field is dangling once the container dies
+
+**Status:** OPEN. Found 2026-09-11 while writing the Dispose-counter gate for
+the spawn-capture leak fix
+(`issues/fixed/spawn-closure-captures-never-dropped-leak.md`). **Pre-existing**
+— reproduces identically under the v0.2.30 seed and under a `develop` tree
+build, with no compiler change of mine in the picture.
+
+## Symptom
+
+```rust
+sink := S.new();                       // S :: atomic(ref(struct(...)))
+{
+  holder := H(tag : i32(7), sink : sink);
+  _t := holder.tag;
+};                                     // holder dies here
+sink.count();                          // SIGSEGV — sink was freed with holder
+```
+
+`rc=139`, faulting inside the `sink.count()` load. `sink` is a live local in
+the enclosing scope and is never consumed by anything the reader can see.
+
+## Reproducer
+
+`issues/repros/local-stored-in-a-struct-field-is-dangling-after-the-container-dies.yo`
+(no threads, no async, no `Dispose` impl needed).
+
+```
+$ yo compile issues/repros/local-stored-in-a-struct-field-is-dangling-after-the-container-dies.yo \
+    --std-path ./std --optimize 2 -o /tmp/repro && /tmp/repro
+rc=139
+```
+
+## What the emitted C shows
+
+```c
+__yo_t0* sink = yo_id_7890();                          // rc = 1
+{ // begin block
+  __yo_t2* holder = __yo_new___yo_t2(7, sink);         // NO __yo_incr_rc on sink
+  ...
+  __yo_decr_rc_atomic((void*)(holder));                // -> 0 -> dispose -> free
+} // end begin block
+yo_id_7894(sink);                                      // USE AFTER FREE
+```
+
+and the container's generated dispose does release the field, correctly:
+
+```c
+static inline void yo_id_7904(__yo_t2* self) {
+  __yo_t0* __yo_disp_f1 = self->sink; // Destructuring sink
+  __yo_decr_rc_atomic((void*)(__yo_disp_f1));
+}
+```
+
+So the store into the field took NO reference while the container's death
+gives one back. `sink` has no scope-end drop of its own either, so the total
+count balances — it is the LIFETIME that is wrong: main's reference was
+silently transferred to a container that dies first.
+
+This is the dup/drop pair optimizer's cancellation
+(`_optimize_dup_drop_pairs`, `src/evaluator/exprs/begin.yo`; the mechanism is
+described in `AGENTS.md` — "a named local passed to a struct literal gets a
+deferred `___dup` (copy semantics), and the move you see in the emitted C is
+manufactured by the dup/drop pair optimizer cancelling that dup against the
+scope-end drop"). Cancelling is sound only when the container lives at least
+as long as the local. Here the container is in an INNER scope, so the dup that
+was cancelled belonged to a shorter lifetime than the drop it was cancelled
+against.
+
+## Why it was not noticed
+
+**The GC masks it for a plain `ref`.** With `S :: ref(struct(...))` the same
+program prints the right answer and exits 0 — not because the accounting is
+right, but because a plain ref holding a ref is CYCLE-CAPABLE, so it is
+GC-TRACKED and `__yo_decr_rc` routes to `__yo_decr_rc_tracked`, which buffers a
+possible root instead of freeing immediately. An `atomic(ref(...))` is an Iso
+type, never tracked, so `__yo_decr_rc_atomic` frees on the spot and the fault
+is immediate.
+
+Measured, same program, only the type constructor changed:
+
+| container / field type | result |
+| --- | --- |
+| `atomic(ref(struct(...)))` | `rc=139`, SIGSEGV in the later read |
+| `ref(struct(...))` | prints the right value, `rc=0` (GC deferral) |
+| `atomic(...)`, container in the SAME scope | prints the right value, `rc=0` |
+| `atomic(...)` + a `Dispose` impl | `rc=139` — `Dispose` is not the trigger |
+
+So the non-atomic half of the language has the same wrong accounting and is
+only saved by a collector that has to run eventually.
+
+## Severity
+
+A use-after-free on a shape a reader writes without thinking: build a handle,
+put it in a struct inside a block, keep using the handle afterwards. It is
+silent at compile time and, on the `ref` path, silent at run time too.
+
+## Fix direction
+
+`_optimize_dup_drop_pairs` must not cancel a dup emitted for a store into a
+container whose scope is INNER relative to the local's own scope — the pair it
+matches has to be same-scope. Per `AGENTS.md`'s warning about this family, the
+change needs the dup/drop emit-diff gate (per-function dup/drop counts before
+and after): a mispaired drop in the other direction is a double-free rather
+than a leak. An over-cancellation canary belongs beside it — the same-scope
+case above must keep its cancellation, or every struct store costs a
+dup/drop pair again.
+
+Tests: the reproducer as a `tests/` case in both spellings (`atomic(ref(...))`
+and `ref(...)`), with a Dispose counter as the oracle rather than a leak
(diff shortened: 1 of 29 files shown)


shd101wyy:
**Local suite: 4056 passed, 0 failed** — `yo test ./tests --exclude tests/internal --exclude tests/cli-cases`, run with the binary this branch builds.

That number is the one that matters for this PR, because the first attempt was **not** clean. With fix 2 (the spawn wrapper's capture release) and *without* 2b (the dup/drop cancellation guard), the same run reported two failures:

```
  ✗ Test Arc shared across thread
  ✗ Test Arc of struct shared across thread
```

both reading `0` instead of `42` after `join()` — the capture struct had been holding a borrowed alias, and releasing it freed the caller's value. Their three-thread sibling passed throughout, for the wrong reason: two capture dups is more than one, and the optimizer only ever cancelled a single dup. 2b is that finding.

Also green with this binary:

| | |
| --- | --- |
| `tests/arc.test.yo` | 15/15 |
| `tests/thread.test.yo` | 13/13 |
| `tests/internal/lexer.test.yo` | 56/56 |
| `yo fmt --check ./src ./std ./tests` | clean |

Reproducer oracles, all with the built binary:

```
after join: shared = 42                                       # arc capture
disposes (must be 1) = 1                                      # spawn Dispose counter
plain scope: closed after the sender's scope ended = true     # channel auto-close
spawn capture: closed after join                   = true     #   ...across Thread.spawn
flat, awaited:       task=2 global=2                          # module global
suspending, awaited: task=2 global=2
suspending, awaited: task=4 global=4
suspending, spawned: task=2 global=2
```

`gates_fast.sh` (GATE 0–8) and `fixpoint_only.sh` are running now; I will post their verdicts here.

shd101wyy:
Local gates, with the binary this branch builds:

| gate | result |
| --- | --- |
| `yo test ./tests --exclude tests/internal --exclude tests/cli-cases` | **4056 passed / 0 failed** |
| `scripts/bootstrap/gates_fast.sh` GATE 0–8 | **`=== T1_DONE (cfix) failures=0 ===`** (embedded-Yo sites=7 failures=0) |
| `scripts/bootstrap/fixpoint_only.sh` | **`FIXPOINT_HOLDS`** — stage2 hollow=0, `0 real (17 string-literal floor, 2 abort-stub)`, STAGE2_RC=0, CLANG_RC=0, STAGE3_RC=0 |
| `yo fmt --check ./src ./std ./tests` | clean |

The fixpoint matters more than usual here: 2b changes RC emission across the whole tree (a closure with a single RC'd capture now keeps both its dup and its scope-end drop), and the stage-2/stage-3 self-compile still reaches a byte-identical fixed point.

PR #564, opened 2026-09-11
verifier: V5 second row — forall_val/exists_val/==> (ghost-only) + inout two-state pinned

shd101wyy:
## Summary

V5 tasks 1 + 4's inout half (`FORMAL_VERIFICATION.md` §6): the quantifier vocabulary real specs need, plus two-state `inout` pinned. **Naming per review (2026-09-11), both maintainer decisions:**

- **`forall` is the plain word** — the lexer had reserved it for verification quantifiers since FORALL_TO_GENERIC; this is what it was held back for, so the hold-back is released (`∀`/`∃` stay rejected with an ASCII hint).
- **`exists` is taken as a keyword**: `std/fs/file.yo`'s path check is defined `_exists` and keeps its public surface via `export(exists : _exists)`; the evaluator's `_evaluate_exists_or_call` dispatches the quantifier **only in ghost context** — outside it, a bare `exists(...)` call goes through the ordinary call path, so every existing `{ exists }` import (`src/lock_file`, `skills_command`, `doc_command`, the tests) keeps working unchanged. The vestigial `BK_EXISTS` constant (zero dispatch sites since the TS era) is deleted.

The rest:

- **`==>` is a new three-char operator** (the only one) — matched ahead of the greedy two-char split; reserved-operator list + GRAMMAR tables (en+zh) updated.
- **`forall`/`exists`**: N binder args `(name : Type)` + the predicate body. The evaluator binds binder names to unknowns of their annotated types in a scoped frame (the type comes from the binder's TYPE argument — the annotation call is a binding form, never evaluated as an expression).
- **Ghost-only gating**: the new `is_ghost_context` EvalContext flag (set at the contract-marker bracket, in `ghost_fn` bodies, and in `evaluated_for_verify`); anywhere else the builtins are a compile error — reaching the verifier report through #557's def-eval chain.
- **The walk** lowers them to `VcTerm.Quant` (raw-spelled binders, shadowing saved/restored; `_collect_declared`/`_rename_vars` are binder-aware) and `ImpOp`; instantiation relies on z3 MBQI (`:pattern` triggers deferred until a benchmark needs them — recorded in the plan).
- **inout two-state** needed no new machinery — pinned by the `inout_bump` proves / `inout_nochange` REFUTES twins.

## Validation

- `tests/internal/verifier_quantifiers.test.yo`: proofs/refutations z3-gated; ghost gate and encoding pins solver-free.
- Local end-to-end vs z3 5.1.0 on the merged tree: `quantifier_forall` PROVES (`(not (forall ((k (_ BitVec 32))) (=> ...)))` → unsat), `quantifier_false` REFUTES with a model, both inout twins behave, the ghost gate fires, and loop_invariant/loop_break/two_state_old still prove. `yo check` green on `std/fs/file.yo` and on an `exists` importer (`src/lock_file.yo`).

## Review follow-ups in this branch

- The earlier `issues/single-expression-fn-body-type-error.md` is **deleted** — mis-filed: `{ x }` without semicolons is a struct value by design (E0007's rule); the fn-body spellings are `(x)` and `({ return x; })\). No bug.
- Merged the latest develop twice (the maintainer's branch-side merge + locally): the waker campaign (#561/#576/#580/#586 — #580 is the await-on-a-struct-field release fix that was breaking every PR's waker test), the build import resolution (#581/#583), and the std campaign.


shd101wyy:
Heads-up on the one red check: **the tier-1 cheatsheet golden diffs are a develop-side regression, not from this branch.** Develop's own latest run ([34627500774](https://github.com/shd101wyy/Yo/actions/runs/34627500774), post-#594/#595/#591) fails the SAME six cases — `init`, `init-build-test`, `init-cwd`, `init-existing`, `skills-install`, `skills-install-zh` — all `content-differs` on the installed `syntax-cheatsheet.md`, with a checkout whose source hash matches the recorded golden (`fddc41…` == `expected_tree`). A seed-binary install copies byte-faithfully (verified locally), so the self-hosted binary's copy path is what changes the bytes. This branch additionally re-records those goldens around the cheatsheet's updated `forall`/`exists` section (e7f17fa8f), but the underlying diff will keep failing every PR's tier-1 until the copy regression itself is fixed on develop.

PR #566, opened 2026-09-11
docs: record what this campaign's PRs cost to learn (codegen, cli-cases, the dup/drop lifetime rule, the waker)

shd101wyy:
Six lessons from landing #550/#554/#555/#556/#558/#559/#560/#561/#562, each one a red gate or a wrong-looking error message that took real time to diagnose. No behaviour change; documentation and six re-recorded goldens.

## `.github/instructions/c-codegen.instructions.md`

Two more rules for emitted-C backtick literals, beside the two already there:

- **A markdown backtick inside a C COMMENT closes the literal.** The error is `E0008: paren-less function and operator calls are not supported` anchored on a word inside a C comment — nonsense until you notice the delimiter, and `yo fmt` makes it worse by padding the fragment to `` ` state ` ``. Hit three times in one session.
- **A new runtime function must be DECLARED before anything that calls it.** The declaration buffer is flushed before the code buffer, so a call emitted into declarations that reaches a later `static` definition is a hard C error, not a warning. Both the waker runtime (hoisted thread-local state + forward declarations into the flags block) and the spawn wrapper's capture drop (`__yo_decr_rc_atomic` / `__yo_incr_rc_atomic` forward-declared in `generate_object_constructor_declarations`) had to be taught this.

## `.github/instructions/testing.instructions.md`

Four cli-case traps, all of which cost a full CI round trip to find:

- **A `cmd` file with no trailing newline runs NOTHING.** The harness's `while IFS= read -r line` loop exits immediately, `stdout.raw` is 0 bytes, and the report says `stdout_keep_match matched nothing — vacuous` — which reads like a wrong regex, so the natural response is to loosen the pattern. Diagnose with `--keep` and `ls -la $W/run/stdout.raw`.
- **A LEXER-level error cannot live in a fixture at all.** CI's `fmt --check` scans the fixture trees and `yo fmt` must lex a file to print it back, so a fixture whose whole point is that it does not lex fails the formatting gate. Those go in `tests/internal/lexer.test.yo` instead — with the offending source written using a DOUBLED backslash, so the test file's own lexing is not the thing under test.
- **`lsp-member-definition` pins absolute LINE NUMBERS inside `std/collections/array_list.yo`**, so any edit that adds lines above `push`, `new` or the type itself moves it. #559's doc sweep did exactly that.
- **Re-recording the skills goldens with a stage-1 binary staged outside the checkout needs `YO_SKILLS` exported**, or `yo skills install` finds no bundled tree and `--record` writes the FAILURE as the golden — rc 1 and ~120 deleted lines instead of one changed hash. `gates_fast.sh` exports it; a hand-run must too. (Found by doing it wrong in this very PR.)

## `AGENTS.md`

The dup/drop pair optimizer's missing lifetime condition, beside the entry that already describes the optimizer: **cancelling a pair is sound only when the container OUTLIVES the local.** A closure definition's deferred dups are its capture-STRUCT field initializers and must never be cancelled — fixed in #562. An ordinary container in an INNER scope has the same unsoundness and is still open (`issues/a-local-stored-in-a-struct-field-is-dangling-after-the-container-dies.md`), masked on the non-atomic path because a cycle-capable `ref` is GC-tracked and its free is deferred.

## `.github/skills/yo-async-effects/async-effects-recipes.md`

`Waker`, `Park`, `park(register, io)` and `yield_now` (#561), with the three properties anything built on them depends on — an early wake is not lost, waking is idempotent, an unwakeable park is reported rather than hung — and the `while(!h.is_finished(), io.await(yield(io), io))` anti-pattern they exist to replace.

## Goldens

The six cli-cases a `.github/skills/` edit moves are re-recorded. The diff is exactly one hash line per case; the full corpus scores **PASS 96 / GOLDEN-DIFF 0 / NO-GOLDEN 0** (1 network case skipped).


shd101wyy:
Corrected after review: **the backtick rule was wrong.**

I had written "never write a markdown backtick inside a C comment in that literal". `\`` is a supported escape in a Yo template literal, so markdown in a code comment is fine — you escape it, exactly as `\$` keeps a literal `${`. Verified rather than assumed:

```rust
c := `// reads \`state\` first, then registers
static int x = 1;`;
```

compiles; the running program prints a real backtick; the emitted C contains a real backtick; and `yo fmt` preserves the `\`` and is idempotent over it. **No lexer or parser bug.**

What stays on record is the failure mode when you forget the escape: the backtick closes the literal, the rest of the C is parsed as Yo, and the error lands as `E0008` on a word inside a C comment — which is why it is worth a line at all.

shd101wyy:
Following up on the backtick question, because my first answer conflated two things.

**Your example is already valid, with no escape.** The lexer skips comments, so this compiles and runs today:

```rust
// This is `valid` without escape — a plain Yo comment
/// A doc comment with `markdown` backticks and a ```rust fence
```

Verified by compiling and running it. No bug, nothing to change.

**The case that bit me is not a comment.** `src/codegen/**` emits the C runtime from backtick TEMPLATE LITERALS, so a C comment written there is *string content*, and `` ` `` is that string's delimiter — an unescaped one closes the literal and the rest of the C gets parsed as Yo. `` \` `` escapes it, round-trips through `yo fmt`, and emits a real backtick.

The instruction now says which case is which, rather than implying backticks are unwelcome in comments.

**And since "you cannot write a backtick there" should not be documented as if it were true**, this PR now also carries `plans/backlog/RAW_AND_FENCED_STRING_LITERALS.md`, which writes up the real fix in two steps:

1. **The diagnostic, which is most of the value and needs no language change.** The lexer knows a token came from inside a template it had been scanning, so `E0008` can add *"this may be inside a backtick template closed early by an unescaped ` on line N; write \\` for a literal backtick"*. That turns a 20-minute diagnosis into a 20-second one.
2. **A triple-backtick fence**, only if step 1 is not enough. Rust's `r#"…"#` is the wrong shape here because it is raw in the sense of *no escapes at all*, and every emitter uses `${…}` interpolation; a `` ``` `` fence keeps interpolation, needs no escape for single backticks, reads like a markdown fence, and is a contained lexer change (the template scanner already tracks a delimiter — it gains a fence length).

Measured so the priority is honest: 65 files in `src/codegen/` emit C from templates, 9774 backticks between them, and exactly **10** escaped ones. Ten escape sites is not a crisis. The error you get when you forget one is.

PR #580, opened 2026-09-11
codegen: awaiting a future held in a struct field must not release it

shd101wyy:
## The failure this fixes

develop's full battery (run [34557965609](https://github.com/shd101wyy/Yo/actions/runs/34557965609)) is RED, and so are every open PR's test legs — all on the same test:

```
✗ Test a wake before the park suspends is not lost
  ==3965==ERROR: AddressSanitizer: heap-use-after-free on address 0x507000003824
  READ of size 1 at 0x507000003824 thread T1
      #0 __yo_decr_rc
      #1 yo_id_17469          <- Park's dispose, dropping self._future
      #4 __yo_user_main
  freed by thread T1 here:
      #2 __yo_waker_release   <- the Waker's dispose
  previously allocated by thread T1 here:
      #2 __yo_async_park_start
```

`test (ubuntu-latest)`, `test (ubuntu-24.04-arm)` and `test (macos-latest)` all report it.

## Root cause

`sm->await_future_N` **owns** its reference: the state machine `__yo_decr_rc`s that slot when the await's result is extracted, when the awaited future aborts, and in its dispose function. Every other way a future reaches that slot already pays for it —

- a **produced** future (an `io.async(…)` block, a `__yo_async_*_start()` extern, a call returning `Impl(Future)`) hands over the reference it just created, and its temp's deferred drop is aliased onto the slot rather than emitted twice (`state_machine.yo`, Phase 1b);
- a **named** future is never stored in the slot at all — the existing comment at the dispatch-mode branch says why: "storing it into the slot would hand the slot an unowned reference (the extraction decref would over-release it)".

A future read out of a **field** was the case nobody had covered. `Park.wait` awaits `self._future`, and the store was

```c
sm->await_future_0 = (void*)(sm->__capture.self->_future);
```

with no `__yo_incr_rc`, while `Park`'s dispose still drops `_future`. Two releases, one reference: the park allocates the future (rc 1), `p.waker()` takes one (rc 2), the await releases one it never took (rc 1), dropping the waker frees it, and `p`'s scope-end drop reads freed memory.

## The fix

All seven await-future store sites now go through `emit_await_future_store`, which takes a reference when the awaited expression is a borrowed place — a field, or a chain of them, rooted at an atom. The predicate is shaped after `arg_is_rc_projection_place` (`src/evaluator/utils.yo`), which asks the same borrow question about an ordinary call argument; the codegen form is syntactic because there is no environment to resolve the root binding in.

```c
sm->await_future_0 = (void*)(sm->__capture.self->_future);
// Borrowed future (read out of a field): the slot releases what it
// holds, so it needs a reference of its own — otherwise this await
// frees the field's future out from under its owner.
if (sm->await_future_0 != NULL) { __yo_incr_rc((void*)sm->await_future_0); }
```

## Verified

- **Emit A/B on the reproducer**: old compiler `incr_slot=0 decr_slot=3`, new `incr_slot=1 decr_slot=3` — one store, one balancing dup, unchanged drops.
- **Emit A/B on `src/main.yo`**: nothing in `src/` awaits a bare field, so the compiler's own emitted C is unchanged (`__yo_incr_rc`/`__yo_decr_rc` counts identical).
- `yo fmt --check` clean; `yo check ./src` 270/270.
- `tests/async/waker.test.yo` 9/9 with the rebuilt compiler.

Note the bug does **not** reproduce by running the program on a dev machine — the freed object's memory has not been reused yet, and the local `--sanitize address` build does not instrument. The emitted C, and the ASan-armed CI legs, are the oracles.

## Also here

- `tests/async/waker.test.yo` — "Test awaiting the same park twice does not release its future twice". Two awaits drive the count to zero while the park and its waker are still live, so the release is unambiguous rather than a one-off imbalance at scope end.
- `issues/fixed/awaiting-a-future-held-in-a-struct-field-releases-it-twice.md` — the record.
- `.github/instructions/c-codegen.instructions.md` — the slot-ownership rule, so a new store site does not reintroduce it.

### DIFF
--- a/.github/instructions/c-codegen.instructions.md
+++ b/.github/instructions/c-codegen.instructions.md
@@ -149,9 +149,10 @@ a bare `sm->await_future_N = …` at a new one. The missing dup was a
 use-after-free in `Park.wait`, which awaits `self._future`
 (`issues/fixed/awaiting-a-future-held-in-a-struct-field-releases-it-twice.md`);
 it does not reproduce by running the program locally (this box's
-`--sanitize address` does not instrument), and only the ASan-armed CI legs —
-both Linux ones and macOS — report it. An RC change around awaits is not clean
-until those legs have run; until then, read the counts in the emitted C.
+`--sanitize address` is inert — the runner prints "AddressSanitizer is not
+functional with this compiler setup … Skipping sanitizer"), while ALL SIX CI
+`test (…)` legs die on it. An RC change around awaits is not clean until those
+legs have run; until then, read the counts in the emitted C.
 
 ## `ExprInfo.variable_name` is UNTRUSTWORTHY in cond/match arm-value position
 
--- a/.github/instructions/c-codegen.instructions.md
+++ b/.github/instructions/c-codegen.instructions.md
@@ -127,6 +127,32 @@ no-ops; `windows-11-arm` gets the latest from choco). An "arm64-only" C failure
 may be a clang VERSION difference, not an architecture one — check the
 versions in the job logs first.
 
+## `sm->await_future_N` OWNS its reference — a borrowed future must be dup'd
+
+The state machine `__yo_decr_rc`s that slot in three places: when the await's
+result is extracted, when the awaited future aborts, and in the state machine's
+dispose function. So whatever is stored there must be a reference the slot owns:
+
+- a future the awaited expression **produces** (an `io.async(…)` block, a
+  `__yo_async_*_start()` extern, a call returning `Impl(Future)`) hands over the
+  reference it just created, and its temp's deferred drop is aliased onto the
+  slot rather than emitted separately (`state_machine.yo`, Phase 1b);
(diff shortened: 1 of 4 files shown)


shd101wyy:
Emit A/B on `src/main.yo` finished: the two emitted `.c` files are **byte-identical** (165,050,613 bytes each; `__yo_incr_rc` 16506, `__yo_decr_rc` 388473 on both). Nothing in `src/` awaits a bare field, so the compiler's own C does not move — the bootstrap fixpoint cannot be disturbed by this change.

`tests/async` is 71/71 with the rebuilt compiler; the full fast suite is running.

shd101wyy:
Local full fast suite with the rebuilt compiler: **4132 passed / 4132 total, 0 failed** (`yo test ./tests --exclude tests/internal --exclude tests/cli-cases`).

PR #586, opened 2026-09-11
codegen: only a CONCRETE type can be trace-monomorphized — and the async Channel over wakers

shd101wyy:
Unblocks stage 3b of `plans/backlog/WAKER_BASED_SCHEDULING.md` by fixing the compiler defect that blocked it, and lands the `Channel` rewrite that has been sitting in `issues/patches/` since this morning.

## The defect

`collect_trace_methods_from_generic_impls` force-specializes the `trace` method of every reference-semantics type in the codegen type registry — `trace` has no call site, so nothing else would. It took **every** registered ref struct/enum as a candidate, and a GENERIC instance can be in that registry: `is_registerable` deliberately admits a struct whose SomeType content sits behind function-typed fields.

Monomorphizing a type that is not monomorphic is meaningless. The impl's own `T` is bound to itself, the body is re-evaluated with the element type unresolved, and the first thing it does with an element fails — in a file the triggering change never touched:

```
error: No matching call found with arguments: (base.add)(i)
    --> std/collections/array_list.yo:862:30
    |
862 |             tracer.visit(base.add(i));
```

Instrumenting the collector shows the two concrete instantiations either side of it specializing perfectly, and shows why the async `Mutex` — same `ArrayList(Waker)` field, same park — never hit it:

```
channel:  ArrayList(Waker) → T = Waker   ✓
          ArrayList(u8)    → T = u8      ✓
          ArrayList(T)     → T = T       ✗   generic instance
mutex:    ArrayList(u8), ArrayList(Waker) only
```

Nothing is lost by skipping the generic entry: the traversal generators look a trace function up by the CONCRETE container's `type_key`, so such a monomorphization could never be called.

## What that unblocks

`std/async/channel.yo`'s `send`/`recv` now park on a waiter queue instead of re-checking on a **1 ms timer tick**, which put a millisecond floor under every hand-off and capped a producer/consumer pair at ~1000 values/second. The held patch is deleted, its issue moves to `issues/fixed/` **with its prediction corrected** — it blamed a stale substitution of `ArrayList(Waker)`, which is not what happens — and the real cause is recorded beside it with the instrumented evidence.

## Verified

- The reproducer compiles **and runs**; before, it did neither.
- `tests/async` **73/73**.
- Full fast suite **4136/4136**.
- `fmt --check` clean.

Red-first: with the same std patch and an unguarded compiler, the `array_list.yo:862` error above is raised.


shd101wyy:
Emit A/B on `src/main.yo` (same source, old vs new compiler): the only difference is **two comment lines** in the emitted C runtime — a `plans/backlog/WAKER_BASED_SCHEDULING.md` → `plans/WAKER_BASED_SCHEDULING.md` path string that came from a plan-doc rename checked out when the probe binary was built, not from this change. 163,199,579 vs 163,199,563 bytes, 16 bytes apart, no code difference.

So the guard removes no trace monomorphization the compiler's own build was using — consistent with the claim that a generic instance's `trace` can never be called (the traversal generators key on the concrete container's `type_key`).

PR #598, opened 2026-09-11
codegen: a static-dispatch call takes its C return type from the CALLEE, not the call site

shd101wyy:
## The bug

A call whose callee resolves through `impl_closure_call_map` is emitted as a
STATIC dispatch — `closure_<fid>(&<capture>, args…)` — to a function whose
prototype codegen has already written. Three emitters decided whether that call
has a **value** by reading the call expression's own type instead, and all three
were wrong the same way.

The call expression's type is the CALLER's view, and it still carries the
unresolved generic whenever the callee specialised that generic away. At
`T = unit` the callee's prototype says `void` while the expression spells the
erasure `void*`:

```c
static inline void closure_yo_id_10905(void* closure_context, int32_t k);
…
void* _file____priv_temp_16750 = closure_yo_id_10905(&(((__yo_t8*)closure_context)->cb), k);
```

`error: initializing 'void *' with an expression of incompatible type 'void'`,
for a program `yo check` accepts.

Every other instantiation is fine — at `T = i32` the call site's own resolution
kicks in and both sides say `int32_t`. It is exactly the ZST that falls between
the two channels: `other_fn_call.yo` deliberately does not take `ei.ty` when it
is unit (a unit expression type also means "value discarded", so it is not
evidence about the callee), and `resolve_some_type_to_concrete` has nothing to
walk to, because the registry that would hold a `generic(T : Type)` binder's
resolution is deliberately not durable (`unregister_some_resolved_concrete`).

Fixing only the call site moves the error one level up, to `init_assignment.yo`'s
binding spill — which already carries the guard for this shape (#`issues/fixed/binding-a-unit-returning-call-emits-invalid-c.md`) and misses it for the same
reason, then emits `void* t = ;`.

## The fix

All three ask the callee. `declarations.yo` grows two helpers:

- `emitted_return_type_string(func_id, context)` reproduces the prototype's
  `override ?? signature` choice from the same three inputs — the registered func
  type, the entry's body, the effect-record-member body suppression.
  `generate_function_declaration` now goes through the same extracted
  `_return_type_override`, so the two cannot drift.
- `static_dispatch_callee_return_type(func_expr, context)` resolves a call's
  callee through the closure map and answers with the above.

**This cannot regress a working program**: the helper answers `Some("void")` only
where the old code emitted `<erasure> t = <void-returning call>` — C that never
compiled.

## Measured

| | result |
| --- | --- |
| `issues/repros/closure-call-void-result.yo` before | 1 C error |
| after call-site fix only | 1 C error, one level up (`void* t = ;`) |
| after | **compiles, runs, prints `ok`** |
| full D18b repro before | 2 C errors |
| after | **1** — and the unit argument now renders as the dead byte `0` the callee's signature expects |

`yo check ./src` 270/270.

## Tests

Red-first: `tests/closure_param_forwarding.test.yo`, "Test a captured generic
callback at T = unit" fails on develop's compiler with exactly
`initializing 'void *' with an expression of incompatible type 'void'`, and
passes here. It asserts more than "it compiles" — the callback must still have
RUN exactly once with the forwarded argument, because the fix turns a
declaration into a statement and a statement dropped rather than emitted would
be a silent miscompile.

## What this unblocks, and what it does not

This is the first of the three walls `plans/STD_API_STABILIZATION.md` records
against **D18b** (`Thread(T).spawn` carrying its result, `join() -> T`). The
plan entry is corrected: the `void*` half was never `parallelism.yo`. Still
open, both filed:

- `issues/generic-channel-send-specialisation-is-called-but-never-emitted.md` —
  `Channel(unit).send` is called and never emitted; its mangled name appears
  exactly once in the whole translation unit.
- `_capture_judgement_type` resolving a captured closure to its capture STRUCT
  and rejecting it as not `Send`.

### DIFF
--- a/issues/fixed/closure-call-binds-a-void-result-to-a-void-pointer-temp.md
+++ b/issues/fixed/closure-call-binds-a-void-result-to-a-void-pointer-temp.md
@@ -1,0 +1,158 @@
+# A call to a captured generic closure binds a `void` result to a `void*` temp
+
+**Found**: 2026-09-12, working D18b (`Thread(T).spawn` carrying its result,
+`plans/STD_API_STABILIZATION.md` §2 D18). **Class**: a codegen call site and
+the callee's own prototype disagreeing about the C return type. **Status**:
+FIXED 2026-09-12 — red-first regression test
+`tests/closure_param_forwarding.test.yo`, "Test a captured generic callback at
+T = unit".
+
+## The error
+
+```
+error: initializing 'void *' with an expression of incompatible type 'void'
+```
+
+## Minimal reproducer
+
+`issues/repros/closure-call-void-result.yo`:
+
+```rust
+_call1 :: (fn(f : Impl(Fn(n : i32) -> unit)) -> unit)({
+  f(i32(1));
+});
+_run :: (fn(generic(T : Type), cb : Impl(Fn(n : i32) -> T)) -> unit)({
+  _call1(((k : i32) => {
+    _v := cb(k);
+    ()
+  }));
+});
+main :: (fn() -> unit)({
+  _run(((n : i32) => ()));
+  println(`ok`);
+});
+```
+
+`yo compile … --optimize 2` emits one C error. The two lines that disagree:
+
+```c
+static inline void closure_yo_id_10905(void* closure_context, int32_t k);
+…
+void* _file____priv_temp_16750 = closure_yo_id_10905(&(((__yo_t8*)closure_context)->cb), k);
+```
+
+The prototype says `void`. The call site says `void*`.
+
+## Why they disagree
+
+Both spellings are computed, from different inputs, by code that never
+compares them.
+
+The **prototype** is written by `generate_function_declaration`
+(`src/codegen/functions/declarations.yo`): the return type is an
+`override ?? get_type_string(result)` choice, and for a closure whose
+signature result is a `SomeT` the override is the BODY's concrete type. The
+closure `(n : i32) => ()` has body type `unit`, so the prototype is `void`.
+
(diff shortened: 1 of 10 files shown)


shd101wyy:
**Local bootstrap fixpoint: HOLDS.**

```
STAGE2_RC=0
0 real (17 string-literal floor, 2 abort-stub) — stage2.c
stage2 hollow=0
CLANG_RC=0
STAGE3_RC=0
FIXPOINT_HOLDS
```

Worth recording why this was worth checking by hand rather than leaving to CI:
`emitted_return_type_string` reaches `_async_override_return_type` and
`get_type_string`, both of which REGISTER things — the async override registers
`type_key(fn.ret) → <block>_state_t*`, and `get_type_string` mints `__yo_tN`
names on demand. Calling them from a CALL SITE during body emission could have
moved type-collection order, and type-collection order is exactly what the
fixpoint gate is sensitive to. It did not.

Also worth recording because it cost me a cycle: my FIRST fixpoint run reported
`FIXPOINT_BROKEN` with 476k differing lines, all of it `__yo_tN` renumbering.
That was an artifact, not a finding — I had rebased the branch onto develop
AFTER building the binary, so stage 2 was emitted by a binary that predated
#594's 45 discard-binding rewrites while stage 3 was emitted from the rebased
tree. A fixpoint run is only meaningful when the binary was built from the
tree it is compiling; rebuild after any rebase before believing the verdict.

Sanity sweep on the same binary: `closure_param_forwarding` +
`closure` + `fn` — 25 passed, 0 failed. The reproducer compiles and prints `ok`.

PR #614, opened 2026-09-12
evaluator: per-definition dependency tracking, steps 1+2 (Phase 3)

shd101wyy:
Implements steps 1–2 of §6 in `plans/INCREMENTAL_COMPILATION_ZIG_LESSONS.md` — the per-definition dependency foundation plus its first user-visible win. Follows #540 (Phase 0), #541 (Phase 1), #603 (Phase 2, stable names).

## What lands

- **Per-definition source hashes** (`src/evaluator/context.yo`): every deferred definition carries `src_hash` = FNV-1a of its canonical `ast_expr_to_string` rendering — comment- and whitespace-insensitive by construction. Keys are `<module>#<index>#<name>`, deliberately **row-free**, so inserting a comment or blank line shifts neither keys nor hashes. The ordered-statement sequence (imports, pragmas, globals) hashes separately per module.
- **Dependency edges at every resolution point**: intra-module forces (`force_pending_def`), already-Done serves (`resolve_pending_definition` — the miss only fires once), and cross-module member reads (property access on a source-namespace value + every destructured import label). The enclosing definition gains a def-level edge; an ordered statement (empty forcing stack) lands in a per-module ordered-read table — the conservative file-level fallback §6 step 5 prescribes.
- **`mm_definitions_changed`** (`src/module_manager.yo`): a parse-only diff — no evaluation — of a changed file against its last walk's records (stmt-sequence hash, def-shape, per-def hashes).
- **`check --watch` fast path** (`src/check_watch.yo`): the round consults the differ first; a comment-only or whitespace-only edit now costs the ~5 ms diff and nothing else.

## Measured

`src/token.yo` probe (hub), watch round after the edit:

| edit | before | after |
| --- | --- | --- |
| comment-only | 222 files re-checked, 375 s | **0 files, 5 ms** |
| whitespace/reformat-only | 222 files, 375 s | **0 files, 5 ms** |
| a real definition change | 222 files, 375 s | unchanged (file-level fallback — correct; per-def invalidation is the remaining §6 step 4) |

Fixture validation (two-file project, edit lands mid-run via `--watch-once`): semantic edit → both files re-checked OK; comment-only → 0 re-checked; **type-breaking edit → both re-checked and FAILED** (rule-5 diagnostic coverage intact).

## Tests

`tests/internal/check_watch.test.yo` grows differ-level gates (identical content, comment/whitespace-only, body edit, ordered-statement edit — each asserted via `mm_definitions_changed`) and an end-to-end comment-only round. Both tests pass (`YO_TEST_LEAK_VERDICT=0` on WSL2 — see below).

Fast suite: 4149 passed / 2 failed — the two are environment tests (`temp_dir returns /tmp on Linux`, `DateTime.now honours TZ`), failing identically without these changes; the ~615 local LeakSanitizer verdicts are the documented WSL2 artifact (see the Phase 2 PR).

## Bug surfaced and filed (not fixed here)

`issues/enum-pattern-bool-payload-not-compared.md`: a boolean-literal payload in an enum match pattern is **bound, not compared** — `.Some(false)` matches `.Some(true)` (3-line reproducer inside; pre-existing on develop). The watch round hit it directly: the differ's `.Some(true)` verdict matched a `.Some(false)` test, silently skipping a genuinely changed file. The round uses the `.Some(v) => !(v)` shape meanwhile; the pattern-matching fix gets its own PR with a failing test first.

## Notes for reviewers

- The hooks are pure table inserts — no evaluation-order, emission, or naming change (Phase 2's gates — self-compile, fixpoint — are unaffected by construction; CI confirms).
- Debug channel: `YO_DEBUG_P3DIFF=1` prints the differ's verdict and the round's skip/drop decision (house style of `YO_DEBUG_LAZY`/`YO_DEBUG_CAPTURE`).


shd101wyy:
**The `test (windows-latest)` / `test (windows-11-arm)` failures are pre-existing on develop — proven, not assumed.**

- Draft PR #622 (a comment-only source change on develop's tip, nothing else) fails `test (windows-latest)` at the **same assertion**: `plain arm awaits once, looping arm polls to completion — a 1ms deadline beats the 5ms work`.
- Why nobody saw it: #608's and #610's merge pushes were classified docs-only and **skipped the test matrix** (15 skipped / 3 success on the d6d910f0 run), so the current develop tip had never run the Windows legs until this PR.
- This PR's emitted C for that test differs from the pre-#608 emission only in #608's cancellation/timer-free-yield machinery — zero hunks from this PR's diff (verified by diffing the batch C against a Phase-2-only compiler).
- Root cause + fix direction: `issues/windows-1ms-deadline-race-loses-since-cancellation-landing.md` — after #608 removed the 1 ms timer under `yield`, the poll loop spins in microseconds while the Windows runtime's clock (`GetTickCount64`, ~15.6 ms granularity) collapses the 1 ms and 5 ms dues into one timer batch; the test's tie-break then awards the first-spawned work task. Fix = QPC-derived `now` in the Windows runtime.

Everything else on this PR is green (26 pass at last count, including the fixpoint, the chunked behavioural gate, the self-hosted tier-1 gates after the golden re-record). I'm treating the Windows timer fix as its own PR against develop rather than expanding this one's scope.

shd101wyy:
**Status of the remaining red checks — all three are develop-side, each with direct evidence:**

1. ~~QPC Windows timer regression~~ — FIXED by #627 (admin-merged after local gates: 201/201 + 208/208 + 76/76).
2. `build-test-suite-flags` golden + 4 × tests/internal shards — **#626's merge left develop's own tree broken**: it deleted `BuildArtifact.runtime_files` but left `tests/internal/build_runner.test.yo`'s `_mk_artifact` passing it (E0605: expected 22, got 23). Fixed on this branch in 89383eba (verified locally 12/12). Worth noting the tier-1 golden diff I chased earlier was MY test error — scoring with a pre-#626 binary; with the correct post-merge binary the golden PASSES unchanged.
3. `test (windows-11-arm)` — a closure-spec pointer-type mismatch that MSVC escalates to an error: the failing batch's emitted C is **byte-identical** between a develop-tip compiler and this branch's (md5-matched), so it is pre-existing on develop and merely composition-dependent (`issues/swallowed-closure-spec-emits-wrong-typed-return-msvc-error.md`). Develop's own next source-touching PR will hit it too.

The branch itself: full local gates green on the post-merge tree (270+/270+ check, the watch fixtures, the CLI case, the internal build_runner suite).

PR #619, opened 2026-09-12
evaluator: a generic function returning Impl(Future(T)) resolves its return PER CALL

shd101wyy:
Two instantiations of `fn(generic(T), v : T, io : Io) -> Impl(Future(T, Io))` in
one program did not compile. Either alone always did. Pre-existing on v0.2.31.

**Stacked on #612** — retarget to `develop` when that merges.

### The evaluator was computing the right answer and discarding it

`YO_DEBUG_RRE=1` on the reproducer:

```
[rre] callee=wrap old=true hkt=Impl : (Future[Future](i32) Io : Io)  resolved_ret=Impl : (Future[Future](T) Io : Io)
[rre] callee=wrap old=true hkt=Impl : (Future[Future](Pair) Io : Io) resolved_ret=Impl : (Future[Future](T) Io : Io)
```

`hkt` is the return-type expression re-evaluated per call, and it is right both
times. The adoption gate threw it away: it accepted a re-evaluated type only
when it was SomeT-free, or when every SomeT in it resolved through its own cell.
An `Impl(...)` existential IS a SomeT with an empty cell, so an async return
could never pass either arm. Every caller then shared one unresolved return, the
await site rendered it `void*`, and the second instantiation's C disagreed with
the first's.

The gate gains a third arm: **adopt when the re-evaluated type no longer
mentions any of this callee's own forall binders** — substitution is then
complete for this call, and what remains are existentials rather than type
variables. Gated on `rre_old_wanted`, which is what keeps the recorded adoption
hazard out: the per-call closure-`F` family leaves a substituted return whose
SomeTs all resolve concretely, so the re-eval is never wanted there at all.

### The binder test compares IDs, not names — and that is not a detail

The first cut compared names, and produced a compiler in which the same program
compiled or not according to what its type variable was called:

| receiver forall | method forall | result |
| --- | --- | --- |
| `U` | `R` | compiles |
| `A` | `B` | compiles |
| `T` | `Q` | **miscompiles** |
| `U` | `E` | **miscompiles** |

`T` and `E` are the names the prelude's own `Future(T, E)` declaration uses, and
those binders ride along inside the re-evaluated `Impl(Future(i32) Io)`.
`_rre_binder_ids` reads the ids off the declared return instead, applying the
name filter exactly once, where the names are unambiguously the callee's.

### Gates

- `tests/generic_future_return_two_t.test.yo` — red-first: fails to compile on
  v0.2.31 ("Cannot unify incompatible types"), 3/3 here. Its second test names
  the forall `T` on purpose.
- `yo test ./tests --exclude tests/internal --exclude tests/cli-cases` —
  **4161 passed, 0 failed**.
- `yo check ./src` — 270/270.
- Bootstrap fixpoint — running; will report.

### Still open, and a different defect

The CLOSURE-PARAM shape (`f : Impl(Fn() -> R)` rather than `v : R`) still
miscompiles: the async block inside the generic function is emitted ONCE for
both instantiations (`conflicting types for closure_yo_id_…`). The return
stamping is no longer the problem there. Measured in
`issues/a-generic-function-returning-impl-future-t-miscompiles-at-a-second-t.md`
with three reproducers and a table of nine variables that are NOT the trigger.
It is what keeps `spawn_blocking` (waker step 5) unlanded.

### DIFF
--- a/issues/a-generic-function-returning-impl-future-t-miscompiles-at-a-second-t.md
+++ b/issues/a-generic-function-returning-impl-future-t-miscompiles-at-a-second-t.md
@@ -1,0 +1,128 @@
+# A generic function returning `Impl(Future(T))` miscompiles at a second `T`
+
+**Status:** the VALUE-PARAM shape is FIXED 2026-09-12; the CLOSURE-PARAM shape
+is still open (see "What is still open").
+**Found:** 2026-09-12, writing `spawn_blocking` for waker step 5
+(`plans/WAKER_BASED_SCHEDULING.md`).
+**Pre-existing:** reproduces identically on the published **v0.2.31**.
+
+## Symptom
+
+Three different C errors depending on the shape, all the same root: ONE
+specialization's view of `T` is used where another's belongs.
+
+| shape | error |
+| --- | --- |
+| `fn(generic(T), v : T, io : Io) -> Impl(Future(T, Io))`, async body `v` | the AWAIT temp is `void*` (T never resolved) while the sync future's `result` is `int32_t` / the struct |
+| same with a closure param `f : Impl(Fn() -> T)` | `conflicting types for closure_yo_id_…` — ONE async-block closure emitted, returning `void*`, with a `Pair`-typed body |
+| an `io.await` inside the async body | `sm->result` typed `int32_t` in the specialization whose `T` is a struct |
+
+Either instantiation ALONE compiles and runs. The pair does not.
+
+## Reproducers
+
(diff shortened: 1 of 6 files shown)


shd101wyy:
Bootstrap fixpoint: **FIXPOINT_HOLDS**, stage2 hollow=0.

(The first run reported BROKEN and was an artifact of my own recorded trap — `$S1` had been built from a different branch's tree, so stage 2 carried a runtime block stage 3's compiler does not emit. Rebuilt from this tree and re-ran; the diff is empty.)

PR #623, opened 2026-09-12
ci: a stacked PR runs a reduced battery, not the whole thing

shd101wyy:
A pull request whose base is not `develop` is a stacked PR. Its diff is only
meaningful against its base, the base's run has already covered everything below
it, and the retarget-to-develop that must happen before it can merge fires a
full run of its own. Running ~28 jobs on every push to every level of a stack
costs hours of runner time and delays the runs that actually decide a merge — on
2026-09-12 a stacked run sat **queued for forty minutes** while develop-targeted
work waited behind it.

### What a stacked PR runs now

| job | why it stays |
| --- | --- |
| `changes` | classifies the diff and now also decides the battery |
| `test (ubuntu-latest)` | the **ASan** leg — it caught a heap-use-after-free on 2026-09-12 that macOS ran green |
| `bootstrap-self-test` (tier-1 gates) | its GATE 0 caught a reproducer left behind by an API sweep the same day — and it is also where the **CLI golden corpus** runs (`gates_fast.sh` → `cli-diff-test.sh` over `tests/cli-cases/`), including `async-blocking-await-inside-task`, whose `opts` set `YO_ASYNC_STRICT=1`. That case is the only gate that fails on an event loop nested from inside a task; without it such a bug does not fail anything, it just makes a parallel scheduler run serially |

Everything else waits for a develop-targeted run: the 6-platform native matrix,
cross-emit, both wasm legs, both fixpoint legs, the hollow sweep, TSan, the musl
static build, the self-hosted internal tests, formal verification, the docs site
and the VS Code extension. The `test` matrix also drops its arm leg on a stacked
PR and keeps the ASan one.

### Why the decision lives in a job output, not an event filter

`on.pull_request.branches: [develop]` would be one line, and the existing
comment on the `pull_request` trigger says why it is the wrong line: an event
filter is how a stacked PR silently skips CI **entirely** the day a naming
convention changes. A `changes` job output is visible in a log line that says
which battery ran and why, and it degrades to "reduced", never to "none".

Anything that is not a pull request — a push to develop, a merge queue entry —
always gets the full battery. Nothing can merge into develop on the reduced one:
branch protection requires all fifteen checks, and on a stacked PR they are
simply absent.

`actionlint` is clean apart from two pre-existing SC2086 notes that are also on
develop.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,13 +4,16 @@ on:
   push:
     branches:
       - develop
-  # No base filter: EVERY pull request gets CI, whatever branch it targets —
+  # No base filter: EVERY pull request gets a run, whatever branch it targets —
   # stacked PRs included, so integration problems surface while the stack is
   # open instead of at the end of the merge train (base-merge → retarget →
-  # rebase → force-push). A base-branch glob here is how a stacked PR
-  # silently skips CI the day the naming convention changes. Branch pushes
-  # alone still run nothing (the push trigger stays develop-only); a PR's
-  # synchronize event covers its branch pushes.
+  # rebase → force-push). A base-branch glob HERE is how a stacked PR silently
+  # skips CI the day the naming convention changes, which is why the scope
+  # decision lives in the `changes` job instead: a stacked PR runs a REDUCED
+  # battery (see its `full` output) rather than none, so it still has signal
+  # and the choice is visible in a job log rather than in an event filter.
+  # Branch pushes alone still run nothing (the push trigger stays
+  # develop-only); a PR's synchronize event covers its branch pushes.
   pull_request:
   # Merge queue support, currently INERT and deliberately kept.
   #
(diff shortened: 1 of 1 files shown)


shd101wyy:
Checked a caveat yo-d1 raised: does the reduced battery still run the CLI golden corpus?

**Yes, and that is a third reason to keep the tier-1 job.** The corpus is not a job of its own — `scripts/bootstrap/gates_fast.sh` runs `scripts/cli-diff-test.sh` over every case in `tests/cli-cases/` (GATE 7/7b, absorbed), and `gates_fast.sh` IS the `bootstrap-self-test` job the reduced battery keeps. So `async-blocking-await-inside-task` — whose `opts` set `env=YO_ASYNC_STRICT=1`, and which is the only gate that fails on an event loop nested from inside a task — still runs on a stacked PR.

That case is worth naming because it is invisible to everything else: without `YO_ASYNC_STRICT=1` a nested loop does not fail, it just makes a parallel scheduler run serially, and `check`, the codegen suites, the internal tests and the stage2/stage3 fixpoint were all green over exactly that on p1/yo-toml-manifest tonight.

Updated the PR body's table to say so.

PR #654, opened 2026-09-12
build: un-vendor markdown_yo — the compiler resolves its own dependency

shd101wyy:
The dogfooding milestone of `plans/BUILD_AND_DEPENDENCY_SYSTEM_REDESIGN.md`, and the end-to-end proof for P1.3/P1.4 on the compiler's own build. **Unblocked**: `SEED_VERSION` reached `v0.2.32` on `develop` (6943c89af), the first published seed that understands `yo.toml`.

## What changes

`src/doc/render_html.yo` and `scripts/build_site.yo` imported the Markdown renderer by submodule path. They now say `import("markdown_yo")`, resolved through the repo-root `yo.toml` and the content-addressed store. The `vendor/markdown_yo` submodule and its `.gitmodules` entry are gone. The manifest, the resolver, `yo.lock` v2 with integrity, the store and the import-root closure all now carry the compiler itself.

**Pinned by `rev`, not by a range.** The commit that compiles against the current language is the head of `migrate-to-latest-yo`, which carries no release tag, so `yo.toml` names the sha the submodule held (`a46f7004…`, `v0.0.4-9-ga46f700`). Un-vendoring therefore changes the resolution *mechanism* and nothing about which source is compiled — deliberately, so a regression here can only be the package manager's.

## Why `--imports` was not the answer

The plan drafted this as "hand the seed `--imports markdown_yo=<store path>` from the bootstrap scripts". That is not sufficient on its own: `--imports` tells a compiler where a dependency **is**, and nothing in the v0.2.31 generation can **put** it there — the scripts would have had to `git clone` the dependency themselves, which is re-vendoring in bash and would be deleted at the next seed bump. Waiting for a seed built from this stack removes all of it: `yo.toml` discovery resolves, `yo install` fetches, and no `--imports` plumbing is needed anywhere.

## CI

`yo build` fetches on its own (`install_dependencies` runs before the graph), so a job that only calls `yo build` is unchanged. A job that calls `yo compile` / `check` / `test` directly does need the store populated first: those resolve import roots but do not fetch, and a missing entry is a compile error naming `yo install`.

Rather than reason per job about which commands it happens to use, every job that runs `yo` against this tree gained one step — `.github/actions/install-deps`, a composite running `yo install --locked` (the committed lock is authoritative in CI; a resolution that would change it fails there rather than drifting). Jobs whose compiler is a downloaded artifact rather than a PATH entry pass its path (`suite-cross-emit`, `bootstrap-fixpoint-stage3`, `chunked-gate`). `submodules: recursive` stays on every checkout — `vendor/mimalloc` is still a submodule. `actionlint` findings unchanged: 70 on `develop`, 70 here, and the two sets are identical line-for-line (all pre-existing shellcheck infos).

**Audited per job, not per workflow**, across all five workflows: 19 `install-deps` sites; every other job either only calls `yo build`, runs no `yo` at all (`vscode-extension`, `seed-bundles-cross`, `portable-c`, `publish-release`, `seed-bump`), or runs `yo` outside this tree — `install-scripts.yml` works in `$RUNNER_TEMP`, and `test-native`'s build smoke runs `yo init`, which writes the scratch project its own `yo.toml`, so the manifest walk-up never reaches this one.

## Four defects found before merge, fixed here

All four were introduced by this change. Numbers 1, 2 and 4 came from auditing every job in all five workflows against what each one actually runs; number 3 was found only by CI, and is the one worth reading.

1. **`chunked-gate` had no `install-deps` step.** It downloads the suite candidate and runs `scripts/bootstrap/chunked_gate.sh`, which compiles `src/main.yo` four times against the checkout — and `src/doc/render_html.yo` is in that closure. It is the single job that was missed, so it would have been the one red job on the first `develop` run after the merge, with exactly the ``git dependency "markdown_yo" is not in the store`` error.

2. **`yo.toml` shipped `version = "0.2.31"` into a 0.2.32 tree**, and nothing would ever have written it again — a third version field beside `src/version.yo` (the source of truth since P2.5 blocker B5) and `vscode-extension/package.json`. release.yml's own comment warns that "two independent bumpers is how they drift apart", and this one was born drifted. Set to 0.2.32 and added to the release pipeline the way the extension is: SET from the computed version, staged by the REQUIRED fail-loudly add list, and added to the bump commit's purity-test exclusion so the `bump=none` resume path still recognises a pure bump. The perl is anchored to the `[package]` table because `yo.lock` uses `version` for its format version.

3. **`sudo` moves the Yo cache root — and the error message misdirects.** Four jobs went red on the first green-seed run (`test` on both arches, both wasm legs), all at their first self-emit:

   ```
   error: import("markdown_yo"): git dependency "markdown_yo" is not in the
   store (/root/.cache/yo/store/sha256/325f5e11…) — run `yo install`
   ```

   `install-deps` had *succeeded* in each of those jobs. It runs as `runner` and fills `/home/runner/.cache/yo`; the memory-capped emit runs under `sudo systemd-run --scope`, `sudo` resets `HOME` to `/root`, and `$HOME/.cache/yo` moves with it. Two stores. **The diagnosis is the path, not the text** — the message names the store the *compile* looked in, never the one the *install* filled, and its own advice ("run `yo install`") is wrong here. It reads exactly like a missing `install-deps` step, so the natural fix is a 20th one that changes nothing. Fixed by forwarding `env HOME="$HOME"` at all six `sudo systemd-run` sites that compile `src/main.yo`: test.yml (stage-1, fixpoint stage-2, stage-3), fixpoint-arm64.yml (stage-2, stage-3), and the `build-stage1` composite behind both wasm legs. Note `fixpoint-arm64.yml` runs on a **schedule**, so that one would have gone red on its own later, not on a PR.

4. **`deploy-site.yml` took an unconditional `yo install --locked`.** That workflow's whole purpose is checking out an *arbitrary published tag* and rebuilding that release's site — the salvage lever for a release whose in-run deploy-site was skipped. **No published tag has a `yo.toml`** (`git cat-file -e <tag>:yo.toml` is ABSENT on every one, v0.2.32 included) because the manifest arrives with this change, and `yo install --locked` exits 1 there with `no yo.toml found … Run 'yo init'`. So it would have broken redeploying *every tag that exists*, at exactly the moment you least want a new failure mode. Guarded with `if: hashFiles('yo.toml') != ''` — the only conditional call site, and the same `if:` is what stops the step resolving `./.github/actions/install-deps`, which those trees also do not contain.

## Gates

Re-earned on this head, not carried forward — #638 rewrote `src/codegen/async/runtime_io_*.yo` after the last recorded verdict, so the emitted C differs.

Local, on a compiler built from **this** head by the v0.2.32 seed:

| gate | result |
| --- | --- |
| `yo check ./src` | 275/275 |
| `yo check ./std` | 175/175 |
| `yo fmt --check src std tests scripts` | clean (rc 0 — it prints nothing when clean, so the rc is the gate) |
| CLI corpus | `PASS 133  GOLDEN-DIFF 0  NO-GOLDEN 0  SKIP 1` (total 134) |
| `tests/internal/{build_runner, pkg_config, doc_render_markdown, doc_stability, manifest, lock_file, install_command}` | 12 / 11 / 26 / 7 / 19 / 17 / 19 passed |
| stage2 ≡ stage3 | **FIXPOINT_HOLDS** — stage2 ≡ stage3, byte-identical at 132,901,618 bytes; `STAGE2_RC=0`, `CLANG_RC=0`, `STAGE3_RC=0`, `stage2 hollow=0` (0 real, scored with `scripts/count-transpile-failures.sh`, not a raw grep) |
| `actionlint` | 70 on `develop`, 70 here — same finding set; one pre-existing `SC2046` info shifts its in-script line number (110 → 116) because a comment was added above it |

Additional checks for the things this change actually moves:

- **Cold store round trip.** With `YO_CACHE_DIR` pointed at an empty directory: `yo install --locked` resolves from `yo.lock` (not by re-resolving the rev), `yo.lock` comes back byte-identical, the store address equals the lock's `integrity` exactly, and a `.verified` sibling is written.
- **`sudo` HOME mechanism**, reproduced and then fixed rather than inferred: with `HOME` pointed at an empty directory the tree-built compiler emits the identical `not in the store` error naming the identical store hash (rc=1); with HOME as-is, rc=0.
- **`yo install --locked` with no manifest** exits 1 (`no yo.toml found … Run 'yo init'`) — which is why `deploy-site.yml`'s step is guarded.
- **`test-native`'s build smoke**, which scaffolds into `./tmp/hello-world` *inside* this checkout: `yo init` + `yo build run` under a parent manifest that declares a dependency still builds and runs, and its scratch `YO_CACHE_DIR` comes back **completely empty** — the manifest walk-up stops at the `yo.toml` that `yo init` writes, so the scaffolded project never reaches this one.

## Follow-up, deliberately not taken here

Cargo fetches for `cargo check` as well as `cargo build`; Yo fetches only in `build`. Making `check`/`compile`/`test` materialize a locked-but-absent dependency would delete the CI step above and the "run `yo install`" error with it. It is not in this change because import-root resolution is synchronous and pre-`Io` by construction (`src/manifest.yo` says so at the top), so on-demand fetching there is a real design change — and it would make the LSP reach the network. Worth doing; worth doing as its own decision. Recorded in the plan.


shd101wyy:
**Rebased onto `9be75a063`** (which carries #656 and #657) and fully re-gated. Still a draft — still blocked only on `SEED_VERSION` reaching v0.2.32.

### The blocker, verified rather than asserted

The v0.2.31 seed on this tree:

```
$ ~/.cache/yo/versions/0.2.31/bin/yo check src/doc/render_html.yo --std-path ./std
error: Module not found: tried ".../markdown_yo.yo" and ".../markdown_yo/index.yo"
```

It has no manifest discovery and falls through to path resolution. Hand the *same* seed an `--imports` file naming the store path and it succeeds, rc=0 — so `--imports` works in that generation; what does not exist is anything that can **populate** the store. That is why the bootstrap scripts would have had to `git clone` the dependency themselves, and why the fix is a seed carrying the `yo.toml` stack rather than scaffolding.

### Cold-store verification (the `install-deps` step every job depends on)

`YO_CACHE_DIR` (`src/cache.yo:43`) pointed at an empty directory — a genuinely cold cache:

- `yo install --locked` resolved **from `yo.lock`**, not by re-resolving the rev, and exited 0.
- `yo.lock` byte-identical afterwards (`diff -q` → IDENTICAL): the CI step cannot drift the lock.
- Store address equals the lock's `integrity` exactly — `store/sha256/325f5e11…` ↔ `integrity = "sha256-325f5e11…"`. Content addressing round-trips.
- A `.verified` sibling was written, so later runs trust the tree without re-hashing.
- Then, same cold cache: `yo check src/doc/render_html.yo` → rc=0, and the literal CI invocation `yo compile src/main.yo --optimize 2 --std-path ./std --emit-c --skip-c-compiler` → rc=0, 132 105 698 bytes, `scripts/count-transpile-failures.sh` **0 real**.

(A naive `grep -c "Failed to transpile"` on that emit returns 16, which is under the 17 string-literal floor and means nothing — use the script, not a grep.)

### Interaction with #657, measured

#657 makes `yo doc` read `[package] name`; this PR adds a root `yo.toml` naming the package `yo`. They do not compose into a changed site title, because #657 looks in `base_path` only with no upward walk and `scripts/build_site.yo` runs `yo doc std/`. With a binary carrying both:

```
<title>std — Documentation</title>
<h2>std</h2>
```

175 modules, 2834 items — unchanged.

### Re-gate on the rebased tree

`check ./src` 275/275 · `check ./std` 175/175 · `fmt --check` · CLI corpus **133 PASS / 0 GOLDEN-DIFF / 0 NO-GOLDEN / 1 SKIP** · `tests/internal` build_runner 12, pkg_config 11, doc_render_markdown 26, doc_stability 7, manifest 19, lock_file 17, install_command 19 · **FIXPOINT_HOLDS**, `stage2 hollow=0`.

shd101wyy:
**Rebased onto `0d8937636`** (now including #614's per-definition dependency tracking) and fully re-gated. Still a draft, still blocked only on `SEED_VERSION`.

The rebase had a genuine conflict: #659 and this PR both rewrite the milestone section of `plans/BUILD_AND_DEPENDENCY_SYSTEM_REDESIGN.md`. Resolved by **merging** rather than taking a side — this PR's landed-state text now also carries #659's measurement (the v0.2.31 `Module not found`, the `--imports`-works-but-nothing-can-populate-the-store finding, and the cold-store numbers). The rebase also left the status banner with two contradicting paragraphs ("the dogfooding milestone landed with them" and "one item is written but not yet shipped"); the second is dropped, since this PR is what ships it.

Checked #614 against this PR's premise before rebasing rather than after — `git show 70ccaffc7 -- src/module_manager.yo` contains nothing matching `import_root|manifest|resolve_import|register_manifest`, so its +118 lines there do not touch the manifest import-root path this change depends on.

### Re-gate on the new base

| gate | result |
| --- | --- |
| `yo check ./src` | 275/275 |
| `yo check ./std` | 175/175 |
| `yo fmt --check` | clean |
| CLI corpus | **PASS 133 · GOLDEN-DIFF 0 · NO-GOLDEN 0 · SKIP 1** |
| `tests/internal/build_runner` | 12 passed |
| `tests/internal/pkg_config` | 11 passed |
| `tests/internal/doc_render_markdown` | 26 passed |
| `tests/internal/doc_stability` | 7 passed |
| `tests/internal/manifest` | 19 passed |
| `tests/internal/lock_file` | 17 passed |
| `tests/internal/install_command` | 19 passed |
| stage2/stage3 | **FIXPOINT_HOLDS**, `stage2 hollow=0` |

The corpus is the one that mattered here: #614 moved `tests/cli-cases/check-watch-once/expected_stdout`, and 0 GOLDEN-DIFF confirms the rebase picked that up correctly rather than resurrecting the old golden.

CI runs on this branch are cancelled on sight until the seed bump — the v0.2.31 seed cannot resolve `import("markdown_yo")`, which is the documented reason this is a draft, so a red run here is expected and not a signal.

PR #661, opened 2026-09-13
match: a literal payload in an enum pattern is COMPARED, not bound

shd101wyy:
Fixes `issues/fixed/enum-pattern-bool-payload-not-compared.md` — the bug the Phase 3 watch work tripped over directly (the differ's `.Some(true)` verdict matched a `.Some(false)` test and silently skipped a genuinely changed file).

## The bug, both halves

```rust
match(o, .Some(false) => 200, .Some(true) => 100, .None => 0)
```

- **Evaluator**: `false`/`true`/`42` atoms parse as plain atoms, so the destructure loop bound a variable literally named `false` — arm selection compared only the variant NAME, so `.Some(false)` matched any `.Some(...)`: silent wrong arm, comptime and (via collapse) runtime.
- **Codegen**: same-variant arms each emitted their own `case TAG:` → clang `duplicate case value` when both spellings appeared (a compile error — this is what the new tests tripped first), or first-arm-wins silently when only one literal arm was present.

## The fix

- Evaluator (`src/evaluator/exprs/match.yo`): arm selection for a comptime scrutinee requires every literal parameter to equal the scrutinee's field value (`_pattern_literals_match_scrutinee`); literal parameters bind nothing.
- Codegen (`src/codegen/exprs/match.yo`): same-variant arms share ONE `case` block. Arms with literal payloads are guarded by `if ((scrut).data.<V>.<field> == <literal>)` in source order; an unguarded arm of the same variant is its catch-all; a trailing `break` keeps an unmatched guarded case from falling into the next case's block. `_emit_destructure_binds` skips literal parameters.

Strings/chars in payload positions still fail loudly at the existing `Expected identifier, "_", or labeled pattern` guard — unsupported, not silently wrong; deliberately left.

## Tests

`tests/match_bind_nothing.test.yo` grows 4 cases (bool discrimination, bool with catch-all binder, int payloads, comptime-known scrutinee) — verified RED first as a C duplicate-case compile failure, now 8/8.

Fast suite with a compiler built from this tree: **4190 passed / 2 failed** — the same two environment tests (`temp_dir`, `DateTime TZ`) that fail identically on every baseline run on this box.


shd101wyy:
I audited `match` for #665 (`plans/MATCH_PATTERN_MATCHING.md`) and built a compiler from this branch (`6fe1f8203`, v0.2.31 seed) to check it against the audit's rows. **The core fix works** — `.Some(false) => 200, .Some(true) => 100, .None => 0` at runtime gives `a=200 b=100 c=0`. But I found one regression that I think should block the merge, plus two gaps worth deciding on. All measured on a binary built from this branch, not read off the diff.

## Blocking: a string literal payload becomes a silent wrong answer

The description says:

> Strings/chars in payload positions still fail loudly at the existing `Expected identifier, "_", or labeled pattern` guard — unsupported, not silently wrong; deliberately left.

They don't fail. A string literal **is** an `Atom`, so it sails past that guard, and `TokenKind.StringLit` isn't in the `Bool`/`Integer`/`Float` set the new `_pattern_param_is_literal` / `_cg_pattern_param_is_literal` test. So the arm is treated as a **binding** whose name is the literal's text.

On its own that would be the old behaviour. What makes it a regression is the other half of the PR: same-variant arms now share one `case`, and only arms with a *recognised* literal get an `if (payload == lit)` guard. A string arm gets no guard, so it runs unconditionally and its `break;` ends the case.

```rust
E :: enum(Tag(name : str), Empty);
pick :: (fn(e : E) -> i32)(
  match(e, .Tag("a") => 1, .Tag(n) => 2, .Empty => 0)
);
// pick(E.Tag("zzz"))  ->  1   (expected 2)
```

| compiler | result |
| --- | --- |
| `develop` (`1780b90cb`) | **loud**: `error: duplicate case value '__YO_T15_TAG'` |
| this branch | **silent wrong answer**: compiles clean, returns `1` |

Emitted C for that case on this branch — the literal bound as a variable, and the second arm's code dead after an unconditional `break;`:

```c
  case TAG: {
    __yo_str _u34_a_u34_ = e.data.Tag.name;   // the literal "a", bound as a variable
    __yo_tmp = 1;
    break;                                     // first arm always wins
    __yo_str __yo_pat_n = e.data.Tag.name;     // dead code
    __yo_str n = __yo_pat_n;
    __yo_tmp = 2;
    break;
    break;
  }
```

Trading a compile error for a wrong answer is worse than the bug being fixed, and the grouping is what removes the C compiler's ability to catch it. Suggested fix inside this PR: recognise `StringLit`/`CharLit` in the literal test and throw explicitly there (something like `Pattern "<lit>" is not supported in a payload position yet`) rather than relying on the identifier guard. Real string patterns are P3 of #665.

Filed with a reproducer: `issues/pr661-string-literal-payload-binds-instead-of-comparing.md` and `issues/repros/pr661-string-literal-payload-binds-instead-of-comparing.yo`.

## Gap 1: a literal-guarded arm still counts as covering its variant

`checked_variant_names.push` fires regardless of whether the arm's payload is constrained, so this passes exhaustiveness:

```rust
pick :: (fn(o : Option(bool)) -> i32)(
  match(o, .Some(true) => 1, .None => 0)   // no .Some(_) arm
);
```

`yo check` passes; at runtime a `.Some(false)` enters the case, fails the guard, hits the trailing `break;`, and the result temp is **never assigned**. Measured: returns `1`. Emitted C:

```c
  case SOME: {
    if ((o.data.Some.value == true)) { __yo_tmp = 1; break; }
    break;                    // .Some(false): nothing assigns __yo_tmp
  }
  case NONE: { __yo_tmp = 0; break; break; }
  }
  return __yo_tmp;            // indeterminate
```

The rule: an arm with a refutable sub-pattern shouldn't mark its variant covered unless a same-variant arm with only irrefutable sub-patterns (or `_`) exists. A red-first test would be the snippet above needing to fail check with `Match expression is not exhaustive`.

## Gap 2: the async emitter is untouched

`_generate_match_with_await_impl` (`src/codegen/async/state_code_gen.yo`) still emits one `case TAG:` per arm, so the same source inside `io.async` with an `await` per arm is the duplicate-`case` error this PR fixes for the sync path:

```
error: duplicate case value '__YO_T_17610571303979907226_SOME'
```

Fine to leave for a follow-up (it's P0 in #665), but worth knowing it's still there.

While probing that emitter I also found a pre-existing bug unrelated to this PR: its destructuring loop binds **positional atoms only**, so a labeled or curly pattern in an await-carrying arm binds nothing and the body reads zeroed SM fields (`.Rect({width, height : h})` prints `0` where the positional twin prints `20`). Present on `develop` too — filed as `issues/async-match-arm-labeled-destructure-binds-nothing.md`.

## Minor: `-1` isn't covered

`.Some(-1)` is still rejected — a prefix-minus is a call, not an atom:

```
error: Expected identifier, "_", or labeled pattern (label: variable) for destructuring parameter, got -(1)
```

Reasonable for a stop-gap, just noting the token-kind test shouldn't become the long-term definition of "literal".

---

Net: fix the string case (and ideally gap 1) and this is good to land — it closes two rows of the audit that nothing else does. #665 has the full audit and the plan for replacing the grouping/guard emitter with a general pattern lowering, so this doesn't need to grow beyond the stop-gap.

PR #695, opened 2026-09-15
verifier: unsigned decreases measures compare unsigned

shd101wyy:
Fixes `issues/verifier-decreases-nonneg-is-signed-for-unsigned-measures.md` (moved to `issues/fixed/`).

## What

`decreases-nonneg` and `decreases-step` hardcoded the SIGNED comparisons (`bvsge`/`bvslt`) regardless of the measure's type, so an unsigned measure (u64/usize — the common counter/length shapes) refuted its own entry ground at the high bit (z3 picks n = 2^63 as "negative"). Both emissions now derive their comparison from the measure expression's own type via `_measure_is_signed` (ExprInfo → `_int_is_signed`, defaulting signed like `_binop_of_ctx`):

- entry ground: `bvsge`/`bvuge`
- self + clique-edge step: `bvslt`/`bvult`

## Fixtures

- Task-4 mutual fixtures flip i32+`requires(n >= 0)` → u64 without the bound (the unsigned ground is trivially true; the n=0 wrap is excluded by the branch guard). `verifier_mutual.test.yo` 2/2 with real z3 locally.
- Signed coverage unchanged (`recursion_decreases.yo` i32, `non_decreasing_measure.yo` refutation).
- Plan task-4 banner updated (quirk marked FIXED).

Filed during V6 task 4 as a warm-up independent of task 3 (handover §4.5).


shd101wyy:
Reviewed — the fix is right, and right in the way that matters: it changes **both** emissions from one `_measure_is_signed` predicate. Fixing `decreases-nonneg` alone would have been unsound, since it is the unconditional nonneg obligation that currently keeps the signed `decreases-step` unreachable; doing both together is what avoids opening that hole. I checked there is no third emission site — only 3777 and 4041.

Two follow-ups, neither a blocker:

**1. The negative fixture doesn't gate the step *direction*.** `mutual_recursion_false.yo` passes `n` unchanged, so the obligation is `n <u n` — false for every `n`. But `n <s n` is also false for every `n`, so that fixture would have passed under the old `BvSlt` too. It's a good regression test for "an unchanged measure refutes", it just can't distinguish signed from unsigned.

The case that does: a measure that **increases across 2^63**. `bvslt(2^63, 0)` is TRUE while `bvult(2^63, 0)` is FALSE, so only a growing-unsigned fixture fails if the step comparison regresses to signed. There is no such fixture in `tests/spec/fixtures/negative/` today. Without it the step direction is gated only by the nonneg obligation happening to fire first — which is precisely the coupling that made the partial fix dangerous.

**2. The loop-variant obligations still hardcode signed.** `loop-variant-continue` (vc.yo:1865), `loop-variant-nonneg` (2000) and the havoc step (2061) are unchanged. I checked whether that's a live hole rather than assuming: `loop-variant-nonneg` is emitted unconditionally whenever the loop carries a measure, so an unsigned variant fails entry at 2^63 and the signed step is unreachable — same conservative false alarm as the function case was, not a soundness bug. But it carries the same trap, so all three want to move together. Filed as `issues/verifier-loop-variant-obligations-are-signed-for-unsigned-measures.md` (landing with #696).

The `macos-26-intel` failure on this PR was a runner DNS outage (`DNS lookup failed: example.com`, 1238 passed / 1 failed), not your change — I reran it. That test treats any failure under CI as a backend regression even though `std/net/errors.yo:85` has a distinct `.DNSFailed` variant; written up in `issues/live-tls-test-treats-a-dns-outage-as-a-backend-regression.md`.

PR #706, opened 2026-09-15
std: JsonValue gains an Int arm so integers round-trip exactly

shd101wyy:
The last actionable row of `plans/STD_API_STABILIZATION.md`.

`JsonValue.Number` held an `f64` for every number, so `10000000` rendered as `1e+07`, and any integer above 2^53 silently lost precision. `serde_json` draws the same distinction inside its `Number` type; Yo needs a separate variant because a variant name cannot be overloaded.

## Classification

From the **consumed slice** — a literal with no `.`, `e` or `E` is an integer by the JSON grammar — not by inspecting the parsed double, which would be lossy for exactly the values the variant exists to preserve. An out-of-range literal falls back to the `Number` arm rather than saturating silently.

## Compatibility is deliberate

`is_number` is true for **both** arms and `as_number` widens `Int` to `f64`, so code that only asks "is it numeric" needs no change. `as_i64`/`as_u64`/`_decode_int` gain an exact path with no round trip, and the `f32`/`f64` decoders accept either arm.

Equality is representation-based like serde_json — `Int(1) != Number(1.0)` — because comparing across arms would make equality disagree with `stringify`, and that asymmetry is worse than the surprise.

## The consumer that would have broken silently

`src/lsp/server.yo` matched only `.Number`, with a `_ => usize(0)` catch-all. Once integers parse as `.Int`, every LSP `line`, `character` and `id` would have read **zero** — no compile error, no test failure, just an LSP that mis-positions everything. `comptime_data.yo` had the milder version and now lifts large integers losslessly instead of rounding.

Adding a variant to an enum whose matches carry catch-alls is exactly this hazard; the consumers had to be enumerated by hand, because no checker reports it.

## A trap in the ToJson sweep

`f32`'s `ToJson` body is **byte-identical** to the ten integer ones (`JsonValue.Number(value : f64(self))`). A blanket replace would have quietly turned floats into integers, so they are rewritten per impl block, with an assertion afterwards that exactly one `f64(self)` form survived.

## Verification

- `yo check ./std` 175/175, `yo check ./src` 275/275
- full suite **rc=0, 4275 passed, 0 failed, 0 batch compile errors**
- JSON suite 73 passed (68 before)

The five new tests **assert the positive** — that an integer literal takes the `Int` arm, renders in decimal rather than `%g`, survives past 2^53 exactly, and that `to_json` emits it. Every pre-existing test passes whether or not the split took effect, so without these the change could have been inert and nothing would have said so.


shd101wyy:
Folded in a correction to an issue doc (docs-only PRs are blocked without `--admin`, so this follows the documented practice of folding docs into a code PR).

**I re-diagnosed the defect I filed yesterday, and my first characterization was wrong.** It claimed `ArrayList(Dyn(Trait)).get()` fails to specialize `_ptr.add`. Measured with controls this time:

| case | result |
| --- | --- |
| `ArrayList(Dyn(Shape))` built + `get()` **in the user module** | compiles |
| `ArrayList(AnyError)` built + `get()` **in the user module** | compiles |
| `ArrayList(i32)` returned **from std/error.yo** | compiles |
| `ArrayList(Dyn(ToString))` returned from std/error.yo | **FTT** |
| `ArrayList(Dyn(ToString))` as a std fn **parameter** | **FTT** |

So it is not "cannot be indexed", not specific to `Error`/`SelfTrait`, and not a general cross-module generic problem. It is precisely: **a generic whose type argument is a `Dyn`, crossing a module boundary**. The `_ptr.add` error is a *second* symptom of the same broken instantiation, visible only when the caller also calls `.get()`.

That narrowing is the useful part — it points at instantiation identity and specialization registration across module loads, rather than at the container's method bodies. Renamed to `issues/a-generic-instantiated-over-a-dyn-cannot-cross-a-module-boundary.md`.

Consequence worth stating plainly: **a std API cannot return or accept a collection of trait objects at all** today. That is what blocks `error_chain`, and it is an expressiveness limit rather than a cosmetic one.

PR #741, opened 2026-09-17
thread_local: module-level per-thread storage (evaluator + codegen)

shd101wyy:
`thread_local` is a **modifier on the variable name**, the way `comptime(v) : i32` is on a parameter. It declares a module-level binding with **one instance per thread**, initialized lazily on that thread's first touch. This closes the language gap behind the `rand.thread_rng` row of `plans/STD_API_STABILIZATION.md`.

```rust
(thread_local(rng) : Rng) = Rng.from_entropy();   // typed
thread_local(counter) := i32(0);                   // inferred

counter = (counter + i32(1));                      // reads and writes are ordinary
```

## No parser change was needed — and the spelling is why

Wrapping the NAME means the LHS stays an ordinary `:` pair, or an ordinary atom for `:=` — the same shapes `comptime(v) : T` and `given(x) := v` already use. So both paths are handed something they already handle, and stripping the modifier is the whole of the front end.

> An earlier revision of this PR wrapped the whole *binding* (`thread_local(rng : Rng) = …`). That put the type inside the modifier, matched none of the three existing LHS shapes so it needed a fourth, and had **no `:=` form at all**, because the type was part of the syntax. The current spelling is the user's correction and is strictly smaller.

## Both binding forms, because they are different code paths

| form | evaluator path | how |
|---|---|---|
| `(thread_local(x) : T) = v` | `exprs/assignment.yo` | strip the modifier from the `:` pair's name position |
| `thread_local(x) := v` | `exprs/initialization_assignment.yo` | strip beside the existing `given(name)` unwrap — the same shape |

Each carries the same two checks and the same registry call, and **each is covered by its own test**: a shared path cannot be assumed across two different entry points.

## Evaluator and codegen land together

Neither half is independently testable: with no codegen there is nothing to run, and `yo check` never evaluates bodies. Shipping them apart would buy verification that does not exist.

**Option 1 of the plan**: a thread-local may not hold a reference-counted type. C11 thread-local storage has no destructor hook, so an RC value would leak once per thread — the declaration is refused with a diagnostic naming the restriction rather than leaking silently. Option 3 (keyed storage with a generated destructor) branches on the *same* condition, so this check is the shape it will reuse.

## Three details that are not obvious from the diff

- **The init flag is set BEFORE the initializer runs**, so an initializer that reads the same thread-local sees zeroed storage and terminates instead of recursing until the stack is gone.
- **Accessor bodies are emitted unconditionally**, not from `generate_main_wrapper`, which is executable-only — otherwise a `--static-library` archive would carry declared-but-undefined accessors.
- **The declaration site must not go through the read substitution.** Reads resolve to `(*name__tl_get())`, an lvalue, so the same substitution is valid in write position — that is what lets the rest of codegen keep treating a thread-local as an ordinary module-level global.

Storage class matches what `src/codegen/async/runtime_core.yo` already emits: `__declspec(thread)` on Windows, `_Thread_local` elsewhere, **dropped on wasm**, where there is one thread and an ordinary global is correct. (`get_uses_tls` in `src/codegen/` is *transport-layer* security — an unrelated use of the same three letters, and the first thing that looks reusable here. It is not.)

## Two bugs earlier builds caught, both worth recording

1. **`E0906`** — `(g_thread_locals : HashMap(...)) = ...` is a runtime binding and therefore order-dependent, and it sat below the rollback function that clears it. This is the one class lazy top-level bindings deliberately do not cover, so it is the failure mode to expect whenever a registry global is introduced.
2. **Exactly half the lowering.** Reads were substituted to `(*name__tl_get())` while nothing declared the storage, the flag or the accessor, so clang reported `indirection requires pointer operand ('int' invalid)`. Recognising the head rebound a *local* and left the AST alone, so `anonymous_module.yo` (which records a module-level init only for the right LHS shape) and `_module_init_var_atom` both failed to see a declaration. **Rebinding one variable is not the same as rewriting the tree** — both strip sites now rewrite the AST.

## Verification

Re-run against the corrected spelling, covering both forms. Results are posted as a comment when the run lands; the battery is:

- a spawning program: each thread's own instance, the main thread's unaffected by either spawn, and the `:=` form behaving identically;
- the emitted C, checked for the storage, the init flag and the accessor;
- both rejections **in separate files**, so neither can pass for the other's reason;
- `yo test ./tests/thread.test.yo`, `check ./src`, `check ./std`.

## What this does not do

`std/rand.yo` cannot use the form until a release ships an evaluator that knows it — `thread_rng` stays seed-gated and is a follow-up, not part of this PR.


shd101wyy:
Converting to draft: the declaration syntax is wrong.

`thread_local` should be a **modifier on the variable name**, the way `comptime(v) : i32` works for parameters — not a wrapper around the whole binding:

```rust
(thread_local(x) : i32) = 12;   // not  thread_local(x : i32) = 12;
thread(x) := 12;                 // inferred form
```

Reworking the evaluator recognition and re-verifying before this is reviewable.

shd101wyy:
## Verification — corrected spelling, both forms

Tree-built compiler at `240ae5cb4`.

```
BUILD_RC=0
TL_COMPILE_RC=0   TL_RUN='5'
```

The spawning program: main thread starts at the initializer, two spawns each get their own instance and mutate it, and the main thread's value is unchanged by either. The `:=` form is asserted alongside in the same program.

**Both forms lower independently** — this is the part a shared-path assumption would have missed:

```c
static _Thread_local int32_t counter_m14555577849700717121;              // (thread_local(x) : T) = v
static _Thread_local bool    counter_m14555577849700717121__tl_init = false;
static int32_t*              counter_m14555577849700717121__tl_get(void);
static _Thread_local int32_t inferred_m14555577849700717121;             // thread_local(x) := v
static _Thread_local bool    inferred_m14555577849700717121__tl_init = false;
static int32_t*              inferred_m14555577849700717121__tl_get(void);
```

**Both rejections, in separate files so neither can pass for the other's reason** — and deliberately one per evaluator path, so each path's checks are exercised rather than assumed:

```
REJECT_RC_TYPE_RC=1      (typed form)  "cannot hold a reference-counted type, and "bad" has type String…"
REJECT_NON_MODULE_RC=1   (:= form)     "must be declared at module level. "inner" is declared inside a function body…"
```

```
yo test ./tests/thread.test.yo   17 passed
check ./src  275/275        check ./std  176/176
```

Taking this out of draft.

PR #752, opened 2026-09-17
issues: the old-value save on a non-dropping assignment is dead code — and that is fine

shd101wyy:
**Downgraded from a defect to a NOTE after review. No fix planned.**

Filed from the question *"does `MaybeUninit` for an array allocate on the stack or the heap?"* — it allocates on the **stack**: `MaybeUninit` is a zero-cost newtype over a by-value struct in an ordinary automatic local, with no `malloc` in the function. Reading that C turned up the old-value save, and I initially overstated it.

## The objection was right, and the measurements back it

Saving the old value before an assignment is **by design**. For a reference-counted old value that temp is exactly what the deferred drop reads:

| assignment | the saved temp is | |
|---|---|---|
| `(s : String) = …; s = …` | read **5 times**, feeding `__yo_decr_rc` | necessary |
| `a(i) = i32(1)` | declared once, never read | dead |
| `i = (i + usize(1))` | declared once, never read | dead |

The dead ones are loads from a local whose address the optimizer can see, never read afterwards — clang removes them at `-O2`. **There is no runtime cost to recover.**

The fix I sketched was also disproportionate: it would have to gate where the temp is *decided* (`attach_temp_variable_to_expr`), not where it is emitted, because suppressing the save without the drop emits a reference to an undeclared temp — the exact failure the `sm->` branch beside it was added to fix. That is a change to every assignment in every program, needing a byte-identity corpus diff, to buy nothing.

## What survives, because it is measured and easy to lose

`Array(T, U).default()` writes into `MaybeUninit` storage, so its dead save reads memory before anything writes it. For most `T` that is an *unspecified value* rather than UB, because the storage's address is taken.

`bool` is the exception in principle: `Array(bool, U)` lowers to `bool data[N]` (C `_Bool`), and an indeterminate `_Bool` outside `{0,1}` is a trap representation, which the standard does call UB. It is still a **dead** load that clang deletes, so nothing has been observed and nothing is expected.

If it ever does matter, the cheap fix is local — have `Array(T, U).default()` not read what it is about to write — rather than changing assignment codegen globally.

This PR is the record and the reproducer, nothing more.


shd101wyy:
Closing: not a defect. The old-value save is by design — for a reference-counted old value it is exactly what the deferred drop reads — and where it is dead it is a load from a local the optimizer can see, removed at -O2. Nothing to recover, and the fix would have touched every assignment in the language to buy nothing.

No issue doc is being added; the observation is not worth carrying as a repo file.

PR #792, opened 2026-09-19
evaluator: infer a bare-variant local's enum from a later return(local) or the body tail

shd101wyy:
## What

`plans/backlog/LLM_AUTHORING_AUDIT_2026-09-19.md` §2.2: a `:=` local initialised from a bare variant literal takes its enum type from where the local flows.

```rust
f :: (fn(x : i32) -> Result(i32, E))({
  r := .Ok(x);      // was: error: Failed to infer enum variant type.
  return(r);
});
```

`return(.Ok(x))` has inferred from the function's result type since the def-eval swallow work; binding the value first did not, and the design cheatsheet's `io.async` rule ("return `io.async(...)` directly — an intermediate variable prevents enum variant type inference") is the same hole seen through async.

- **`evaluate_begin_expression`** (`src/evaluator/exprs/begin.yo`): before evaluating a statement of the shape `name := .V` / `name := .V(…)`, look ahead in the same block for the local's first typed use — a `return(name)` anywhere in a later statement, or `name` as the function body's tail expression — and, when the enclosing function's result type is known, store it as a one-shot hint.
- **`evaluate_initialization_assignment`**: consumes the hint (read + clear) as the RHS expected type when — and only when — the RHS is a bare variant literal. Every other RHS is evaluated without an expected type, as before.
- **`EvalContext.local_variant_hint`**: a DEFAULTED field (`?= None`), so the three constructor sites and the internal tests' contexts need no change.
- **`property_access.yo`**: the three "Failed to infer enum variant type." sites carry a `help:` with the repair options (name the enum, annotate the binding, or return the value directly).

Deliberately NOT general bidirectional inference: the variant literal is the only expression whose type is unknowable without context, and the block-local forward look for `return(local)` / the body tail is the smallest rule that covers the reported shapes (the plan's recommendation; the argument-position case is left for a later slice).

## Tests

- `tests/enum_variant_local_inference.test.yo` (5 tests: `return(r)`, a fieldless variant, the body tail, a `return` nested in a branch, the local used before the return) — red on the v0.2.37 seed (`Failed to infer enum variant type`), green on the branch build.
- `tests/cli-cases/check-enum-variant-local-needs-a-type`: the shape with no later use still fails, now with the help text.

## Gates (branch build with the v0.2.37 seed, after rebasing onto develop at #791)

- `check ./src` 275/275; `gates_fast.sh` — see the final comment.


shd101wyy:
Gates on the rebased branch build (v0.2.37 seed): `gates_fast.sh` `T1_DONE failures=0` (repros, battery with hollow detection, corpus goldens, `check ./std`, `check ./src` 275/275, CLI execution differential, fmt, 159 CLI cases incl. the new golden, embedded programs); `tests/enum_variant_local_inference.test.yo` 5/5.

PR #810, opened 2026-09-20
ci probe (no-op): does the internal-tests shard-1 batch C error reproduce on clean develop?

shd101wyy:
Empty commit on a25f908ce. #806's CI failed ONE job — Compiler internal tests shard 1 — with a C-level 'declaration after case label / unexpected type name' codegen error in the batch for tests/internal/evaluator_utils_is_valid_variable_name.test.yo. That file is old and green on every prior battery; the probe answers whether the failure is develop's own (then it belongs to #803/#804's recent evaluator work and I will file it separately) or specific to #806's diff. Draft + will be closed immediately after the shard verdict.


shd101wyy:

PR #812, opened 2026-09-20
ci probe P1 (draft): Variable.is_ghost_binding field alone — no behavior change

shd101wyy:
Isolation probe for #806's shard-1/2 batch-compile failures: this adds ONLY the defaulted Variable field (no gate change, no contracts change). Shard verdict isolates layout/type-id perturbation from gate semantics. Draft — closed after the verdict.


shd101wyy:

PR #813, opened 2026-09-20
ci probe P2 (draft): gate behavior change with contracts/identifer reverted to develop — field + old gate

shd101wyy:
Isolation probe for #806: env.yo field KEPT, contracts/identifer gate REVERTED to develop's name-set. Shard verdict isolates the gate behavior change. Draft — closed after the verdict.


shd101wyy:

PR #827, opened 2026-09-21
plans: SAFE_MODE — no UB and a total failure model in safe code (backlog design)

shd101wyy:
## What

Design plan for **safe mode with no undefined behavior and a total failure model** — written to backlog (nothing implemented). Two files: the new `plans/backlog/SAFE_MODE.md` (539 lines) and its index entry in `plans/README.md`. Docs-only; no code, no std, no gates affected.

## The design in one paragraph

The slogan is split into two promises. **A — no UB:** safe Yo code cannot execute C-level undefined behavior; Yo emits 100% of the C, so every UB source reachable from safe constructs gets a proof or a defined dynamic check (deliverable by phases 0–4). **B — "no runtime exception" read as:** every runtime failure is *proven* (verifier, tier 1), *typed* (`Result`/`Option`/effects, tier 2), or a *deterministic trap with file:line* (tier 3) — never UB, never silent. Banning traps outright is an explicit non-goal (it would ban `assert`/`unwrap`, break std's additive-only window, and bury LLM authors in match boilerplate); a trap-free strict mode is deferred as phase 6 on top of the proof tier.

## Grounding

Every claim in the hazard inventory was verified against the tree (surveyed at `a1df43578`; anchors re-checked, `plans/README.md` conflict-free against the current tip): the four unchecked `->data[idx]`/`.ptr[idx]` emission sites, raw-C int `/`/`%` (incl. `INT_MIN / -1`), overflow/shifts/casts, the **silent escaped-unwind rc=0** (`__yo_effect_escaped` is checked after module init but never after `__yo_user_main`), and `--sanitize undefined` advertised-but-rejected. Also documented: what is already solid (container impls, borrow counter, Send/Acyclic, comptime checks) and the fact that `docs/en-US/MEMORY_SAFETY.md` *already claims* "All indexing is bounds-checked" — phase 1 closes a doc-vs-code gap rather than inventing a promise.

## For the reviewer

- **Two rulings block phase 3** (§13): **D1** overflow policy (recommendation: trap on signed *and* unsigned, all `-O` levels, matching comptime) and **D2** float→int out-of-range cast (recommendation: Rust-style saturating). D3–D6 carry recommendations too.
- §2 non-goals are worth a skim — they encode the refusals: no ambient throw from `/`, no std API break (checks live in codegen builtins), no ban on `arr(i) = va` (the hazard is the unchecked builtin, not the syntax).
- Phases carry grep-acceptance gates (e.g. `->data[` may only appear inside the shared checked-index helper), a perf protocol whose escalation order never weakens a check or makes semantics `-O`-dependent, and the expectation that fixpoint/hollow-sweep will flush latent compiler bugs as traps — each with an `issues/` entry per workflow rules.


shd101wyy:
**Amendment (2026-09-22): D7 adopted — Phase 0c, the class-1 panic ban.** New commit `6b650f6` (+136/−20 in the plan, README bullet updated).

The maintainer asked how we *determine* which panic-like functions safe mode bans. The answer is now the core of Phase 0c — a single classification question, **"where does the failure information live?"**:

- **Class 1 — failure is in the TYPE, the call erases it**: argument is `Option`/`Result`, callee returns the payload and traps the failure case (`unwrap`, `expect`, `unwrap_err`). → **compile error in safe files now.** No verifier needed: the discarded information is already static. Call-site gate only — no std API change.
- **Class 2 — value-level precondition** (`i < len`, `rhs != 0`, rune boundary): tier-3 traps per phases 1–3; banning these *today* would force `Result`-typed operators — a different language — so they wait for Phase 5b/6. (`String.substring` therefore stays legal despite panicking: classification is by the criterion, not by whether a `try_*` twin exists.)
- **Deliberate abort** (`panic`, `assert`): never banned — the author speaking, and the test infrastructure.
- **Environment failure** (OOM, borrow aborts): keep, documented.

Plus: the **comptime carve-out** (unwrap legal wherever decidable — CTFE checks it; under 5b the spelling upgrades to `requires(is_some)`, so the ban is "until checkable", not forever), the **mechanism** (evaluator registry gated at *both* call resolution and method extraction — extraction is a real bypass path), **exemptions** (`*.test.yo`, `std/` on a 234-site ratchet, `AllowUnsafe`, comptime), and **governance** (`public_safe_report` cross-check; std policy *no new class-1 APIs*; the hand-rolled `unwrap2` loophole is accepted as the ban working).

Migration measured: **163 sites / 18 files in `src/`**, 0 in `vendor/`/`scripts/` (so no dependency or GATE-8 churn), `tests/`'s 2,497 sites exempt by rule.

PR #828, opened 2026-09-21
safe-mode 0a: loud effect unwinds (task aborts + top-level belt) and a main-signature fix

shd101wyy:
Phase 0a of `plans/backlog/SAFE_MODE.md` (stacked on #827 — the plan; do not merge before it).

**What landed**

1. **Unwind-side async task aborts are loud.** A fire-and-forget `io.spawn` whose task unwinds through its effect handler used to exit rc=0 with nothing anywhere (`issues/fixed/effect-unwind-escaping-a-task-or-main-is-silent.md`, repro inside). Both unwind-side `-2` writers — `emit_async_future_escape` and the sync_fut resume escape block — now print `unhandled effect unwind aborted an async task` on stderr. Cancellation (`JoinHandle.abort()`, used by `race`/`timeout`) reaches the same state through `__yo_task_abort` **without** an unwind and is deliberately untouched; `JoinHandle.await` still returns `.None` as the typed channel.
2. **The post-`__yo_user_main` flag belt** on all three `generate_main_wrapper` arms (POSIX/Windows/wasm): abort with a message if `__yo_effect_escaped` survives past the program body, mirroring the module-init check.
3. **Surfaced by the repro, fixed here** (`issues/fixed/main-effect-parameters-other-than-io-segfault.md`): `main :: (fn(io : Io, exn : Exception) -> unit)` compiled and **segfaulted** at the first `exn.throw` — the wrapper zero-initializes parameters and only the `Io` record's members are name-dispatched builtins, so the `Exception` record carried a NULL `throw` pointer. `_reject_non_unit_main` now rejects any `main` parameter that is not the nominal `Io` struct; `DESIGN.md` (en+zh) documents the tightened contract.

**Implementation corrected the plan's H9 mechanism** (recorded in the issue and in SAFE_MODE.md): the flag cannot reach top level — every unwind is caught at a task boundary or install site, and the boundaries clear it. So the check in (2) is unreachable through today's surface; the real observable fix is (1). Both are kept.

**Validation** (no self-hosted build exists on this box — the known reaper constraint; CI is the gate):
- the four touched compiler files pass `yo check` individually (rc=0);
- the exact emitted C snippets were validated behaviorally by patching the seed's emitted program for the repro and running: the diagnostic line appears, rc stays 0;
- two cli-cases, goldens **hand-recorded** (the recording binary must carry the fix, so no seed can produce them): `task-effect-unwind-diagnostic` (stderr line, rc 0) and `main-non-io-effect-param-rejected` (compile error, rc 1);
- `yo fmt --check` clean on all touched `.yo`; fixtures fmt-clean for GATE 6.

CI expectations: the new cli-cases run in the tier-1 gate; existing tests that print the new diagnostic (awaited-unwind flows) may show one extra stderr line — `yo test` asserts don't read stderr, so no impact is expected.


shd101wyy:
**Fix pushed (`9e1a3ae0`): both CI failure classes root-caused and addressed.**

1. **The 4 `algebraic_effects` failures were the belt check firing on legitimate programs.** Patched the kept batch C and ran per-test: a handler bound locally and invoked directly at the install scope (`(raise : Raise) = handler; raise(...)`) escapes the frame CORRECTLY but leaves `__yo_effect_escaped` set — every later effect protocol resets the flag in its prologue, so the dirt is invisible until main exit, where my belt aborted. This is a real flag-hygiene defect at the install-frame exit (`_call_is_handler_installation` rule 1 misclassifying in the batch context), not a swallowed error. The belt is removed and DEFERRED to the new issue `issues/effect-install-frame-exit-leaves-the-escaped-flag-dirty.md` (full evidence + re-land recipe); the async-side diagnostics stand (async_await battery: 210 passed, rc=0).

2. **The 6 NO-GOLDEN cases all had `main :: (fn(io : Io, exn : Exception) -> unit)` fixtures** — the exact NULL-handler-parameter shape the new main-signature check rejects. The check was RIGHT; these fixtures are its first real-world catches. Migrated all six to bind `exn` inside main (the DESIGN.md prescription); none throws through it in its golden path, so behavior is identical.

This means the plan's "flag cannot reach top level" claim was ALSO wrong in the sync direction: the flag DOES ride to main dirty on install-frame exits — the corrected story is in the issue and SAFE_MODE §3.

PR #829, opened 2026-09-21
safe-mode 0b: --sanitize undefined is real (UBSan wiring)

shd101wyy:
Phase 0b of `plans/backlog/SAFE_MODE.md` (stacked on #828; do not merge before it).

Closes H11: `yo compile --help` has advertised `address|thread|undefined` (en+zh) since the sanitize flag shipped, but the validation accepted only `address|leak|thread` — `--sanitize undefined` was a documented flag that always errored.

- Validation (`src/main.yo`) accepts `undefined`; the consumption `cond` gains the arm emitting `-fsanitize=undefined -fno-sanitize-recover=all -fno-omit-frame-pointer`. `-fno-sanitize-recover=all` is the load-bearing choice: a recovering UBSan run (print-and-continue) is not a gate; UB must abort.
- UBSan is this campaign's measurement instrument: the phases 1–3 sweeps (self-built compiler + test suite under it) run through this flag.
- New cli-case `sanitize-undefined-accepted`: rc=0 + emitted-C artifact ignored. It deliberately does NOT assert a runtime UBSan trap — sanitizer runtime functionality is per-machine (the repo already probes ASan usability for this reason; the local nix box's UBSan is inert even on a minimal C probe), and the campaign's UBSan sweeps run where the runtime works.

Local validation: `yo fmt --check` clean on `src/main.yo` and the fixture. `yo check src/main.yo` is reaper-killed on this box (~9 min SIGTERM, the documented constraint) — CI is the validator for this file, as for every `src/main.yo` change.


shd101wyy:
Rebased onto the 0a fix (belt deferred, six fixtures migrated). No content change to this PR.

PR #831, opened 2026-09-21
safe-mode 0c: the class-1 panic vocabulary is a compile error in safe files (D7)

shd101wyy:
Phase 0c of `plans/backlog/SAFE_MODE.md` (stacked on #829 ← #828 ← #827; do not merge before them). Implements the DECIDED D7 ruling: `unwrap`/`expect`/`unwrap_err` on `Option`/`Result` in safe files is a compile error.

**The criterion, as implemented.** A call may not erase a failure the type carries. Three gate sites over shared predicates in `memory_safety.yo`:

1. **property_access metadata gate** — the receiver's `ExprInfo` names the type; no double evaluation (variable receivers, the dominant form).
2. **property_access TypeVal arm** — `Option(i32).unwrap` as an *extraction* (binding the method as a callable value), the bypass the plan warned about.
3. **function.yo dispatch gate** beside `_reject_private_method_call` — the receiver type comes from the resolved method's own first parameter, catching fresh call-result receivers (`make(flag).unwrap()`) the metadata gate cannot see yet.

Exemptions: `*.test.yo` (2,497 in-tree sites stay), `std/` (234 sites, ratchet), `pragma(Pragma.AllowUnsafe)`, auto-generated code. **No comptime exemption** — measured on the seed, a comptime-known `.None` still emits a runtime unwrap (rc 134), so the carve-out waits for CTFE folding (plan updated; the assumption failed, recorded).

**Migration corrected by measurement:** 155 of the 163 `src/` sites already sit in `AllowUnsafe` files; the real migration was **19 sites in 5 files** — visible `__yo_panic("invariant")` arms for provably-Some sites, and `Result.Err` returns for the genuinely fallible version parsers in `resolver.yo` (a malformed component is now a parse error naming the input, not a crash — the function exists to report exactly this).

**Validation:** all eight touched files pass `yo check` individually on this box; three cli-cases (call form rejected, extraction form rejected, `AllowUnsafe` allowed) with hand-recorded goldens; docs updated in both languages (`MEMORY_SAFETY.md` §The Optimistic Panic Vocabulary). Deferred deliberately: the E0xxx/`yo fix` registry entry (pointer-gate precedent: plain message + help is the house shape for mode gates), the `public_safe_report` class-1 cross-check section.


shd101wyy:
Rebased onto the 0b/0a fix. No content change to this PR beyond the fixture migrations now living in 0a.

PR #841, opened 2026-09-22
collections/sync: wrap-by-design arithmetic migrated to wrapping_* (mix_u64, atomic fetch_add/sub)

shd101wyy:
Second 3a-ii find from CI: two more intentional-wrap sites my inventory missed, both now breaking under the Phase 3 traps.

1. **`std/collections/hash_map.yo` `mix_u64`** — murmur3's fmix64 finalizer; its own doc comment said "`*` wraps on u64 overflow (as FNV-1a relies on)". The overflow trap aborted every HashMap/HashSet operation at runtime — the walker/module battery failures, `hashmap_overwrite_no_leak`, and closure corpus cases all funnel through here. → `wrapping_mul`.

2. **`std/sync/atomic.yo` CAS-loop `fetch_add`/`fetch_sub`** (9 widths × 2 ops) — C11 `atomic_fetch_add` is DEFINED as two's-complement wrap; the TSan leg's "AtomicI8 wraps like the runtime operator" test aborted with `integer addition overflow (at std/sync/atomic.yo:319:56)`. → `wrapping_add`/`wrapping_sub`. The i32/isize native `atomic_fetch_*_explicit` intrinsics are correct as-is (the hardware op defines the wrap).

The bounded capacity math (`capacity * 2`, load-factor thresholds) stays plain — guarded logic, not wrap-by-design.

**Validation**: every migrated file `yo check` + `fmt --check` clean; `tests/sync` (the TSan leg's corpus) exercises "AtomicI8 wraps like the runtime operator" directly; the collections batteries pin HashMap/HashSet behavior.


shd101wyy:
Third site found by CI and fixed: the **wrap tests' own expected-value computations** in `tests/sync/atomic.test.yo` used plain `+`/`-` at exactly the overflow boundaries — migrated to `wrapping_add`/`wrapping_sub` (they are the by-design-wrap case the methods exist for). The atomic.yo wrapper migration from the previous commit is confirmed working (the wrapper no longer traps; the trap had moved to the test file's own arithmetic).

